PR #2, opened 2020-04-29
WMS-134 | Add new Text component

PedroHLC:
Objective:
* Follow [Sufiyan’s design pattern](https://www.sketch.com/s/a252928f-582c-4f38-a5ee-14d60a2a3067/a/zaaeQy) guidelines
* Follow Elm’s with-pattern API guidelines
* Develop the Text component

PR #3, opened 2020-04-30
WMS-135 | Add new Button component

PedroHLC:
Requires #2 

Objectives:
* Develop the Button component
* Follow Sufiyan’s design pattern guidelines
* Follow Elm’s with-pattern API guidelines
* Make use of #4 
* Also implement `ColorPalette`


docs/src/Buttons.elm
```diff
@@ -1,249 +1,189 @@
...
 
-import Element as El
-import Element.Background as Background
+import Element
 import Msg as Msg
-import UI.Button as B
-import UI.Colors as Colors
-import UI.Icons as Icons
+import UI.Button as Btn
+import UI.Icons as Ico
```
jouderianjr, line 6: I would call this `Icon`

docs/src/Buttons.elm
```diff
@@ -1,249 +1,189 @@
...
 ```elm
--- Disabled
+-- Text
 
-B.button YourMessage 
-    |> B.withPrimaryColor
-    |> B.withDisabledMode True
-    |> B.withIcon (Icons.map "Show Map")
-    |> B.toEl
+Button.bodyText "Some Text"
```
jouderianjr, line 41: I think it's better to have consistency in the example, `Button.*` or `Btn.*`

docs/src/Buttons.elm
```diff
@@ -1,249 +1,189 @@
...
+          ]
         , { note = """
 ```elm
-B.button YourMessage
-    |> B.withText "Click here"
-    |> B.withDangerColor
-    |> B.toEl
+-- Text
+
+Button.bodyText "Some Text"
```
jouderianjr, line 174: Same here (`Button` or `Btn`)

src/UI/Icons.elm
```diff
@@ -30,6 +34,40 @@ import Element exposing (..)
...
+
+-- NEW
+
+
+type Icon
+    = Toggle
+    | Close
+
+
+toggle : Icon
```
jouderianjr, line 49: Why this is called toggle as it renders a map? 

I'm aware that we will have to refactor everything here, but the icons should receive a text as a hint, accessibility purposes  :D

(Mental note, we should rename this component to UI.Icon)

src/UI/Icons.elm
```diff
@@ -30,6 +34,40 @@ import Element exposing (..)
...
+    Toggle
+
+
+close_ : Icon
+close_ =
+    Close
+
+
+toEl : RenderConfig -> Icon -> Element msg
+toEl _ ico =
```
jouderianjr, line 60: Small stuff, I'd prefer to use `icon` instead of `ico`

src/UI/Internal/Palette.elm
```diff
@@ -0,0 +1,110 @@
...
+    Element.Color
+
+
+
+-- Tones
+
+
+type alias ToneColors =
+    { darkest : Color
+    , middle : Color
```
jouderianjr, line 28: maybe `default` instead of `middle`? I think it makes more explicit that this is the "usual" color to be used.

src/UI/Internal/Palette.elm
```diff
@@ -0,0 +1,110 @@
...
+    , light : Color
+    , lighter : Color
+    , lightest : Color
+    }
+
+
+gray : ToneColors
+gray =
+    { darkest = rgb255 14 20 32 -- #0E1420
```
jouderianjr, line 37: I liked the idea of adding the hex color :D can you add in all the colors? I know it's boring but I think it's a good idea

src/UI/Button.elm
```diff
@@ -1,342 +1,427 @@
...
-                    }
-                , hover =
-                    { bg = Theme.white
-                    , text = Theme.black
-                    , border = Theme.warning
-                    }
-                }
-        }
 
+withStyle : ButtonStyle -> Button msg -> Button msg
```
jouderianjr, line 137: I would kill this, as the user should not be able to set an outline button, at least now :D

src/UI/Button.elm
```diff
@@ -1,342 +1,427 @@
...
                 }
 
         _ ->
-            el attrs <|
-                elFromContent options.content
+            Element.el attrs <|
+                elFromBody cfg body
+
+
+baseAttrs : Button msg -> List (Attribute msg)
```
jouderianjr, line 247: Keep the styles when it is a link button.

docs/src/Buttons.elm
```diff
@@ -1,249 +1,189 @@
...
         "Buttons"
-        [ primaryStory
-        , successStory
-        , warningStory
-        , dangerStory
-        , disabledButtonStory
-        , toggleStory
-        , fullWidthStory
-        , linkStory
+        [ primaryStory renderConfig
```
jouderianjr, line 14: - We should remove the outline story
- We should add a functional toggle story, so maybe it's better to remove toggle from all the stories and create a specific toggle story but with a message and model.

src/UI/Button.elm
```diff
@@ -1,342 +1,427 @@
...
+            [ Background.color active
+            , Font.color passive
+            , Element.mouseOver
+                [ Background.color hoverActive
+                , Font.color hoverPassive
+                ]
+            ]
+                ++ Element.colorTransition 100
+
+        ( _, StyleOutlined ) ->
```
jouderianjr, line 316: I think there is a size difference between the outline and the normal button because of the border size. So maybe we should have a border in the normal button as well.

src/UI/Button.elm
```diff
@@ -1,342 +1,427 @@
 module UI.Button exposing
     ( Button
+    , ButtonClick
```
jouderianjr, line 3: We can kill:

- bodyEl
- withStyle

And then, we don't need to expose ButtonStyle.

src/UI/Icons.elm
PedroHLC: I also thought about that, but as it would add a giant diff to the review I left it to be commited at the end

PedroHLC:
@jouderianjr  Maybe `UI.Utils.ARIA` deserves a PR as enhancement after the main components are ready.

PedroHLC:
As bce2b8d9a0ac1c2a65fd1b7871e3bccab3c29064 is the last moddification requested by @jouderianjr in Slack I'll considered this reviewed, and push it so I can work on [WMS-140](https://paacklogistics.atlassian.net/secure/RapidBoard.jspa?rapidView=22&projectKey=WMS&modal=detail&selectedIssue=WMS-140&quickFilter=34).

PR #4, opened 2020-04-30
WMS-136 | Add RenderConfig (Global runtime settings for all components)

PedroHLC:
Requires #2 

What this PR does?:
* Add a property to AppConfig that dictates global details about rendering components.
* It can be used to determine responsive constraints at runtime.
* In the future, it can be used for things like dynamic dark-mode.

Missing:
* Obtain real window's width and height in paack-ui's docs!
* Documents on how to initialize and use.

### DIFF
--- a/src/UI/RenderConfig.elm
+++ b/src/UI/RenderConfig.elm
@@ -1,0 +1,49 @@
+module UI.RenderConfig exposing (RenderConfig, fromWindow, isMobile, isPortrait)
+
+import Element
+
+
+type alias RenderConfigData =
+    { deviceClass : Element.DeviceClass
+    , deviceOrientation : Element.Orientation
+    }
+
+
+type RenderConfig
+    = RenderConfig RenderConfigData
+
+
+fromWindow : { window | height : Int, width : Int } -> RenderConfig
+fromWindow window =
+    let
+        { class, orientation } =
+            Element.classifyDevice window
+    in
+    { deviceClass = class
+    , deviceOrientation = orientation
+    }
+        |> RenderConfig
+
+
+isMobile : RenderConfig -> Bool
+isMobile (RenderConfig { deviceClass }) =
+    -- TODO: Is Tablet mobile?
+    deviceClass == Element.Phone
+
+
+isPortrait : RenderConfig -> Bool
+isPortrait (RenderConfig { deviceOrientation }) =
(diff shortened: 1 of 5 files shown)


src/UI/RenderConfig.elm
```diff
@@ -0,0 +1,49 @@
...
+    deviceClass == Element.Phone
+
+
+isPortrait : RenderConfig -> Bool
+isPortrait (RenderConfig { deviceOrientation }) =
+    deviceOrientation == Element.Portrait
+
+
+
+{- -- Future thoughts:
```
jouderianjr, line 40: Maybe we should put this at the beginning of the file? 🤔

src/UI/RenderConfig.elm
```diff
@@ -0,0 +1,49 @@
...
+    in
+    { deviceClass = class
+    , deviceOrientation = orientation
+    }
+        |> RenderConfig
+
+
+isMobile : RenderConfig -> Bool
+isMobile (RenderConfig { deviceClass }) =
+    -- TODO: Is Tablet mobile?
```
jouderianjr, line 30: That's a good question, let's leave only `Phone`now :D

src/UI/RenderConfig.elm
```diff
@@ -0,0 +1,49 @@
...
+    deviceClass == Element.Phone
+
+
+isPortrait : RenderConfig -> Bool
```
jouderianjr, line 34: I'm not sure if we should support landscape mobile screens 🤔

PedroHLC: I believe that applies for desktop users as well. There are vertical screens, and users who split their screen with two apps. In this last scenario, and as an example, a Navbar would be better than a Sidebar.

PedroHLC:
> It's good to add tests for `UI.RenderConfig`

I do believe that it's better to test it inside components tests. Otherwise, we would be just testing `Element.classifyDevice` itself, an outside component.

src/UI/RenderConfig.elm
PedroHLC: Also, this serves as an example of how to correctly use this API, that is those `is` functions instead of casing through the internals...

jouderianjr: yeah, sounds good :D

PR #7, opened 2020-05-06
WMS-142 | Required changes for login screen (pt 1)

PedroHLC:
This PR:
* Adds RenderConfig.rootAttributes
* Rename Icons to Icon, removing old icons so we can recreate them from design files
* Add color to Texts, with bigger flexibility to internal components (e.g: Button)

Note that Alert and ActionBar gained some TODO and dummy placeholders, those will be ripped out when those components are refactored.


src/UI/Icon.elm
```diff
@@ -0,0 +1,79 @@
...
+close : String -> Icon
+close hint =
+    Icon (Properties hint Close)
+
+
+toEl : RenderConfig -> Icon -> Element msg
+toEl _ (Icon { hint, glyph }) =
+    case glyph of
+        Toggle ->
+            fasIcon "map" hint
```
jouderianjr, line 51: should we call this Map instead of  `Toggle`?

src/UI/RenderConfig.elm
```diff
@@ -44,3 +45,9 @@ isMobile (RenderConfig { deviceClass }) =
 isPortrait : RenderConfig -> Bool
 isPortrait (RenderConfig { deviceOrientation }) =
     deviceOrientation == Element.Portrait
+
+
+rootAttributes : RenderConfig -> List (Attribute msg)
```
jouderianjr, line 50: As `elm-ui` defines the root elemen as `layout` maybe we should call this `layoutAttributes`?

jouderianjr: To be honest, idk if here is the best place to put it, but at the same time, I don't know any place else to put it

src/UI/Text.elm
```diff
@@ -14,251 +15,100 @@ module UI.Text exposing
     , subtitle1
     , subtitle2
     , toEl
+    , withColor
     )
 
 import Element exposing (Attribute, Element)
 import Element.Font as Font
 import List
+import UI.Internal.Text as Internal exposing (..)
```
jouderianjr, line 24: what do you think about not expose all here? I got confused a little to identify where a function comes from e.g `attributes`

src/UI/Text.elm
```diff
@@ -14,251 +15,100 @@ module UI.Text exposing
...
+import UI.Palette as Palette
```
jouderianjr, line 25: Should we move this module to `UI.Internals.Palette` as well? this way it's more explicit that this module is not "exposable/exposed"

src/UI/Icon.elm
PedroHLC: It's just a placeholder for a proper "toggling" icon.
I also believe they should be called by their function instead of what they look like. This way it's the developer's fault if it chooses to have an icon representing a wrong function.

src/UI/RenderConfig.elm
PedroHLC: > maybe we should call this `layoutAttributes`?

I originally did, then I thought "root" was more explicit.
Now considering that it's made to be used with Element's layout and it alone, it feels obvious that it should be called `elLayoutAttributes`.

~I wrote something very very wrong in that comment~

src/UI/Text.elm
PedroHLC: There is a `UI.Internals.Pallete` and a `UI.Palette`

PedroHLC:
@jouderianjr Please review it again, now focusing on the `UI.TextField` API.

src/UI/Icon.elm
jouderianjr: I think in some cases the same icon could be used for different functions, as will be used across different projects

src/UI/TextField.elm
```diff
@@ -0,0 +1,354 @@
+module UI.TextField exposing
```
jouderianjr, line 1: Maybe we should implement this in another pr?

PedroHLC: Yep, moving to another PR...

PR #11, opened 2020-05-12
Adjust text attributes

jouderianjr:


src/UI/Button.elm
```diff
@@ -451,7 +454,7 @@ elFromBody : RenderConfig -> ButtonBody msg -> Element msg
 elFromBody cfg body =
     case body of
         BodyText str ->
-            Text.body1 str
+            Text.subtitle2 str
```
PedroHLC, line 457: If "Text attributes are the same of subtitle1" why are you invoking `subtitle2`?

src/UI/Button.elm
```diff
@@ -217,8 +217,11 @@ toEl cfg ((Button { click, body } _) as btn) =
 
 baseAttrs : Button msg -> List (Attribute msg)
 baseAttrs btn =
+    -- Text attributes are the same of subtitle1
     [ Font.size 16
-    , Font.center
+    , Element.spacing 8
```
PedroHLC, line 222: As Button is an internal component, you can use `Text.attributes cfg Text.SizeSubtitle2 Text.defaultColor`
Though, that shouldn't be necessary as we're using the text component for rendering it :thinking:

src/UI/Internal/Text.elm
```diff
@@ -1,4 +1,4 @@
-module UI.Internal.Text exposing (..)
+module UI.Internal.Text exposing (Options, Properties, Text(..), TextColor(..), TextSize(..), attributes, defaultColor, defaultText, deskAttributes, fontColor, mobileAttributes)
```
PedroHLC, line 1: I oppose that we should care about what an internal module exposes or not, that's the whole concept of having internal functions. They're here to not be used, but if needed there's no limit to what we can internally do.

src/UI/TextField.elm
```diff
@@ -492,8 +492,11 @@ genericAttr label isPlaceholder hasError width =
 
         else
             1
+
+    -- This Font attributes are the same of Subtitle2
     , Font.size 14
```
PedroHLC, line 497: Here too, for better design consistency, in line 477 you can reuse the `UI.Internal.Text.attributes` function to get text attributes.

PR #13, opened 2020-05-14
Update UI.Alert

jouderianjr:
Update code to match the new design system updates.

- I've removed the `withCloseButton` because Icon is not exposing a way to set the color, I will do it in another pr.

### DIFF
--- a/src/UI/Text.elm
+++ b/src/UI/Text.elm
@@ -5,7 +5,7 @@ module UI.Text exposing
     , body2
     , caption
     , colorBgMiddle
-    , colorInherit
+    , colorInverted
     , colorPrimary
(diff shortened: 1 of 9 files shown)


src/UI/Text.elm
```diff
@@ -100,6 +102,16 @@ colorPrimary =
...
 
 
+colorBgMiddle : TextColor
+colorBgMiddle =
+    Internal.ColorBgMiddle
+
+
+colorInherit : TextColor
+colorInherit =
+    Internal.ColorInherit
```
PedroHLC, line 112: You're giving the developer god-like powers :scream_cat: 
iddqd

PR #14, opened 2020-05-14
WMS-163 | Implement SelectList component

PedroHLC:
Objective:
* Follow [Sufiyan’s design pattern](https://www.sketch.com/s/a252928f-582c-4f38-a5ee-14d60a2a3067/a/zaaeQy) guidelines
* Follow Elm’s with-pattern API guidelines
* Develop one component capable to:
  * [X] Offer a selectable list
  * [x] Action bar
  * [x] Search field

Things that got in together:
* Badge component :name_badge: 
* NavigationContainer's Dialog :smile: 

Things that tried to get-in:
* Text's ellipsis (Not implementable with elm-ui's Text or Paragraph, see: https://github.com/mdgriffith/elm-ui/issues/112)


src/UI/Badge.elm
```diff
@@ -1,55 +1,58 @@
-module UI.Badge exposing (dark, error, light, primary, success, warning)
+module UI.Badge exposing (Badge, badge, toEl)
```
jouderianjr, line 1: Why did you remove the badge variations? they are there in the design system, check it [here](https://www.sketch.com/s/a252928f-582c-4f38-a5ee-14d60a2a3067/a/G33Qz8)

src/UI/Badge.elm
```diff
@@ -1,55 +1,58 @@
...
 
+toEl : RenderConfig -> Badge -> Element msg
+toEl cfg (Badge { content } { tone }) =
+    Text.overline content
+        |> Text.toEl cfg
+        |> Element.el
+            [ Element.width shrink
+            , Font.center
+            , Background.color Palette.gray.lighter
+            , Element.paddingEach { top = 4, bottom = 4, left = 5, right = 3 }
```
jouderianjr, line 46: left and right padding should normally have the same number, I checked the sketch inspector and it had 5, so I think you could have `paddingXY 5 4`

src/UI/SelectList.elm
```diff
@@ -0,0 +1,210 @@
...
+import UI.Icon as Icon exposing (Icon)
+import UI.Internal.Palette as Palette
+import UI.RenderConfig exposing (RenderConfig)
+import UI.Text as Text
+import UI.TextField as TextField
+import UI.Utils.ARIA as ARIA
+import UI.Utils.Element as Element
+
+
+type alias SearchOptions object msg =
```
jouderianjr, line 19: We currently don't know how this use case will be, so I would remove it until we understand better what we need to do here.

src/UI/Button.elm
```diff
@@ -245,7 +245,7 @@ buttonPadding ((Button { body } _) as btn) =
             Element.paddingXY 31 15
 
         BodyIcon _ ->
-            Element.paddingXY 9 15
+            Element.paddingXY 9 13
```
jouderianjr, line 248: Why this was changed?

src/UI/Badge.elm
PedroHLC: WIP

src/UI/Badge.elm
PedroHLC: Letter-spacing is moving text 1px to the left on both Chrome and Firefox for me!

src/UI/SelectList.elm
PedroHLC: It's already working, on the spaces screen it got something like this, feedbacks are welcomed:
```elm
viewLoadedSpaces : List Space -> Dialog.State Dialog -> Maybe String -> RenderConfig -> Element Msg
viewLoadedSpaces spaces dialog currentFilter cfg =
    selectList (OpenSpaceDetails >> ForDialogMsg) .id spaceItemView
        |> SelectList.withOptions spaces
        |> SelectList.withSearchField
            "Space Name / Barcode"
            FilterSet
            (Maybe.map (\str -> ( str, spaceHasString )) currentFilter)
        |> SelectList.withActionBar "Create New Space"
            Icon.add
            (UuidGen (ForDialogMsg << OpenSpaceCreation))
        |> SelectList.toEl cfg

-- ...
spaceHasString : String -> Space -> Bool
spaceHasString str { name, barcode } =
    String.contains str name
        || String.contains str barcode
```

src/UI/Button.elm
PedroHLC: Look at docs, icon-buttons were rendering taller than text button, while on Sketch they're actually smaller (48px vs 40px - was gonna ask you about this anyway)

PedroHLC:
> There are a lot of things here are not related to the pr, can you remove it? (Mainly deletions)

Ok, reverted the deletions, but Badges and Dialog (beside WIP) are too small and one-file change only, I think moving them would only slow the development.

src/UI/Button.elm
PedroHLC: Sufyian said 48px of height! This makes it!

PedroHLC:
@jouderianjr, this is back to WIP as I'm writing some tests!

src/UI/NavigationContainer.elm
```diff
@@ -235,3 +291,40 @@ menu applier { menuExpanded } =
...
+        ]
+        [ Element.column
+            [ Element.width (Element.fill |> Element.maximum 640)
+            , Element.centerX
+            , Element.centerY
+            , Background.color <| Palette.gray.lightest
+            ]
+            [ Element.row [ Element.width Element.fill ]
+                [ Text.heading5 title |> Text.toEl cfg |> Element.el [ Element.width Element.fill ]
+                , Icon.close "Close dialog" |> Icon.toEl cfg |> Element.el [ Events.onClick close ]
```
jouderianjr, line 313: I think we should add `pointer` here

src/UI/NavigationContainer.elm
```diff
@@ -235,3 +291,40 @@ menu applier { menuExpanded } =
 lazyMap : (a -> b) -> (c -> a) -> (c -> b)
 lazyMap applier original =
     \whatever -> applier (original whatever)
+
+
+dialogView : RenderConfig -> Dialog msg -> Element msg
```
jouderianjr, line 296: Maybe it would be better to move this to another module? Maybe `Internal.Dialog`?

src/UI/RenderConfig.elm
```diff
@@ -37,6 +37,19 @@ fromWindow window =
         |> RenderConfig
 
 
+mapWindow : { window | height : Int, width : Int } -> RenderConfig -> RenderConfig
```
jouderianjr, line 40: this is not exactly a map, right? maybe we could rename to `updateWindowSize`, `setWindowSize` or `updateWindow`.

src/UI/SelectList.elm
jouderianjr: I said that because we haven't defined if that would be a remote filter, local filter, and so on. it would probably be a remote filter as we will add pagination here. But I'm ok if you prefer to keep this.

src/UI/SelectList.elm
```diff
@@ -0,0 +1,232 @@
...
+    { label : String
+    , searchMsg : String -> msg
+    , currentFilter : Maybe ( String, String -> object -> Bool )
+    }
+
+
+type alias Options object msg =
+    -- FUTURE: sort: Maybe (List object -> List object)
+    { itens : List object
```
jouderianjr, line 27: maybe it's a typo here, would be `items`?

src/UI/SelectList.elm
```diff
@@ -0,0 +1,232 @@
...
+    , searchField : Maybe (SearchOptions object msg)
+    , actionBar : Maybe ( String, String -> Icon, msg ) -- TODO
+    , isSelected : Maybe (object -> Bool)
```
jouderianjr, line 30: I would say that in the middle of the code when I saw `isSelected` I thought it was a `Bool` instead of a function that returns a Bool, so I would suggest calling maybe `isSelectedFn` or something like that because normally we would use `is*` as a Bool, but it's up to you.

src/UI/SelectList.elm
```diff
@@ -0,0 +1,232 @@
...
+itemView : RenderConfig -> Properties object msg -> Bool -> object -> Element msg
+itemView cfg { select, renderItem } isSelected obj =
+    let
+        selectedBg : List (Attribute msg) -> List (Attribute msg)
+        selectedBg =
+            Palette.primary.middle
+                |> Background.color
+                |> (::)
+
+        bgColor =
```
jouderianjr, line 192: So, I'm not a huge fan of point-free style, I think can cause a lot of confusion. mainly because you needed to add a type signature on top to explain what's going on, and we almost never use this inside of `let...in` block.

I think the name of the function is really misleading. a function called, when you think about a function called `bgColor` what it supposes to return?  I would think maybe an `Element.Color` or an `Element.Attribute Msg`, but will never think about returning a list of attributes that are affecting sizing, spacing and etc. and also, it is not select, this function is not even adding a background color.

my suggestion would be

```elm
        selectedBg =
            Background.color Palette.primary.middle
```

and then

```elm
Element.el
    [...attributes
    ]
    |> prependIf isSelected selectedBg

prependIf : Bool -> a -> List a -> List a
prependIf shouldAdd item items =
    if shouldAdd then
         item :: items
    else
         items
```

Ps. this `prependIf` could go in the helper module we discussed before

src/UI/Badge.elm
```diff
@@ -1,55 +1,110 @@
...
+            , Element.height (px 20)
+            , Primitives.roundedBorders
+            ]
+
+
+
+-- Internal
+
+
+toneInside : BadgeTone -> ( Element.Color, TextColor )
```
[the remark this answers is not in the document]
PedroHLC: Just a helper function for getting color scheme, I'll rename it `toneToColors`, cause that's what it does...

src/UI/NavigationContainer.elm
PedroHLC: Yep, agreed!

src/UI/NavigationContainer.elm
PedroHLC: Indeed

src/UI/SelectList.elm
PedroHLC: It is :portugal: fault!

src/UI/SelectList.elm
PedroHLC: I prefer `selected` as a boolean, `isSelected` to `sth -> Bool`. We should start writing contribution guidelines...

src/UI/SelectList.elm
PedroHLC: I liked this `prependIf` a lot, doing it!

PR #17, opened 2020-05-20
Text API v2

PedroHLC:
This takes care of two things that were annoying me:
* Paragraphs were blindly used (they still are misused, but we can fix them now)
* Text was hacking colors palette, now the only one doing it is the light button when it's disable.

This took me more time then I wished for, sorry...

### DIFF
--- a/docs/src/Palette.elm
+++ b/docs/src/Palette.elm
@@ -28,8 +28,8 @@ tones =
     ]
 
 
-brightnesss : List Palette.Brightness
-brightnesss =
+brightnesses : List Palette.Brightness
+brightnesses =
     [ Palette.brightnessDarkest
     , Palette.brightnessMiddle
     , Palette.brightnessLight
@@ -64,7 +64,7 @@ colorVarianceView : ( String, Tone ) -> Element msg
 colorVarianceView ( label, tone ) =
     Element.row
         [ Element.spacing 20 ]
-        [ brightnesss
+        [ brightnesses
             |> List.map (colorView tone)
             |> Element.row []
         , Element.text label
--- a/docs/src/Palette.elm
+++ b/docs/src/Palette.elm
@@ -28,13 +28,13 @@ tones =
     ]
 
 
-lums : List Palette.Brightness
-lums =
-    [ Palette.lumDarkest
-    , Palette.lumMiddle
-    , Palette.lumLight
-    , Palette.lumLighter
-    , Palette.lumLightest
+brightnesss : List Palette.Brightness
+brightnesss =
+    [ Palette.brightnessDarkest
+    , Palette.brightnessMiddle
+    , Palette.brightnessLight
+    , Palette.brightnessLighter
+    , Palette.brightnessLightest
     ]
 
 
@@ -64,7 +64,7 @@ colorVarianceView : ( String, Tone ) -> Element msg
 colorVarianceView ( label, tone ) =
     Element.row
         [ Element.spacing 20 ]
-        [ lums
+        [ brightnesss
             |> List.map (colorView tone)
             |> Element.row []
         , Element.text label
--- a/docs/src/Palette.elm
+++ b/docs/src/Palette.elm
@@ -41,7 +41,7 @@ lums =
 colorView : Tone -> Palette.Brightness -> Element msg
 colorView tone brightness =
     Element.el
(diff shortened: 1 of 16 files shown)


src/UI/Text.elm
```diff
@@ -99,50 +96,42 @@ overline content =
...
+
+multiline : (String -> Text) -> List String -> Text
+multiline style lines =
+    lines
+        |> List.map (style >> Internal.getSpans)
+        |> List.concat
+        |> Internal.Text
+
+
+combination : List Text -> Text
```
jouderianjr, line 132: what is the use case for that?

src/UI/Badge.elm
```diff
@@ -88,23 +92,23 @@ toEl cfg (Badge { content } { tone }) =
...
 
 
-toneToColors : BadgeTone -> ( Element.Color, TextColor )
-toneToColors tone =
+toneToColor : BadgeTone -> Palette.Color
+toneToColor tone =
     case tone of
         ToneLight ->
-            ( Palette.gray.lighter, Text.colorInverted )
+            Palette.color Palette.toneGray Palette.lumLighter
```
jouderianjr, line 99: I think this is more a Palette refactor than a Text refactor haha :D 

That being said, I would strongly suggest refactoring this `Lum/Brightness`part, I think it causes too much confusion, I would suggest using only `brightness`.

For example, why a `lumDarkest` returns a `Brightness`? we should stick to a single term, either `lum` or `brightness`.

src/UI/Palette.elm
```diff
@@ -26,14 +57,25 @@ type Brightness
...
 
 
 toElColor : Color -> Element.Color
-toElColor ( tone, brightness ) =
+toElColor (Color { tone, brightness } { alpha, contrast }) =
     tone
-        |> toColors
+        |> ifThenElse contrast contrastColors toColors
         |> getLum brightness
+        |> colorWithOpacity alpha
```
jouderianjr, line 68: I didn't realize this before, but I'm not sure if this is a good function name, as every color has opacity, so this is more an `updateColorOpacity`, `setColorOpacity` or something like that.

src/UI/Internal/Text.elm
```diff
@@ -276,3 +262,29 @@ mobileAttributes size =
             , Font.letterSpacing 2
             , Font.extraBold
             ]
+
+
+mapOpt : (Options -> Options) -> Text -> Text
```
jouderianjr, line 267: I would call this `mapOptions` or `mapOpts` as this is not mapping an `Option`, this mapping `Options`.

src/UI/Internal/Text.elm
```diff
@@ -276,3 +262,29 @@ mobileAttributes size =
...
+mapOpt applier (Text spans) =
+    spans
+        |> List.map (spanMapOpt applier)
+        |> Text
+
+
+spanMapOpt : (Options -> Options) -> Span -> Span
```
jouderianjr, line 274: same above.

One thing I have to mention, I'm not a huge fan of abbreviations, for example, using `opt` instead of `option`, because in general, I don't see any benefit, but I now that you like a lot, so we have to balance :D (just a thought)

src/UI/Text.elm
PedroHLC: elm-ui allows styling first in text elements, then paragraph elements. But, as we use line-height, our smallest unit is a paragraph. By merging Texts together we can wisely style the text element. This would mainly allow:
```plain
Start some text like this, THEN WOW, THIS IS OVERLINED! But turns back normal! (Small captions at the end).
```

But would also allow us to space text elements differently in the same text, like in Action Bar we have
```
SOME TITLE WITH BODY1 SIZE
Some Caption
```

We can make that every time this happens he have the correct spacing btw them.

src/UI/Badge.elm
PedroHLC: NoooYesss

docs/src/Palette.elm
```diff
@@ -28,20 +28,20 @@ tones =
...
 
 
-lums : List Palette.Brightness
-lums =
-    [ Palette.lumDarkest
-    , Palette.lumMiddle
-    , Palette.lumLight
-    , Palette.lumLighter
-    , Palette.lumLightest
+brightnesss : List Palette.Brightness
```
jouderianjr, line 31: `brightnesses`

PR #19, opened 2020-05-21
WMS-143 | Add Spaces screen (pt 3A)

PedroHLC:
New features:
* New icons
* Badge can have custom brightness
* Buttons can have message mapped
* Alert now has a "light" (gray.lighter) tone
* Dialogs can define a maximum width on its container
* Dialogs are now responsive
* ContentStackChild now has buttons!

Fixes:
* Table width was not being applied
* TextField's label rendering code was totally wrong
* Button.ButtonBody doesn't need that "msg" anymore
* Button.BodyIco renders square
* SelectList search's field doesn't have visible label

### DIFF
--- a/src/UI/Badge.elm
+++ b/src/UI/Badge.elm
@@ -1,4 +1,4 @@
-module UI.Badge exposing (Badge, danger, dark, light, primary, success, toEl, warning)
+module UI.Badge exposing (Badge, danger, dark, light, primary, success, toEl, warning, withAlpha)
 
 import Element exposing (Element, px, shrink)
 import Element.Background as Background
@@ -20,7 +20,9 @@ type alias Properties =
 
 
 type alias Options =
-    { tone : BadgeTone }
+    { tone : BadgeTone
+    , alpha : Float
+    }
 
 
 type BadgeTone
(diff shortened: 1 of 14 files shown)


src/UI/Badge.elm
```diff
@@ -34,54 +36,62 @@ type BadgeTone
...
+    Badge { content = content } { defaultOptions | tone = ToneDanger }
 
 
 success : String -> Badge
 success content =
-    Badge { content = content } { tone = ToneSuccess }
+    Badge { content = content } { defaultOptions | tone = ToneSuccess }
+
+
+withAlpha : Float -> Badge -> Badge
```
jouderianjr, line 67: Screenshot 2020-05-22 at 12 56 24

Sufiyan said that this is in fact darkest primary instead of having  Alpha

src/UI/Internal/Dialog.elm
```diff
@@ -22,12 +24,13 @@ dialogMap : (a -> b) -> Dialog a -> Dialog b
...
     { title = data.title
     , close = applier data.close
+    , width = data.width
     , body = Element.map applier data.body
     }
 
 
 view : RenderConfig -> Dialog msg -> Element msg
-view cfg { title, body, close } =
+view cfg { title, body, close, width } =
```
jouderianjr, line 33: The dialog background is white

jouderianjr:
Screenshot 2020-05-22 at 13 00 54

the close button should be a left arrow right?

jouderianjr:
May-22-2020 13-02-21

Something is not right in the interaction here

PR #21, opened 2020-05-22
WMS-145 | Package screen required changes

PedroHLC:
* [Icon] Add withColor (don't break API with ColorInherit by default);
* [Icon] Add toggle-up/down icons;
* [SelectList] Remove padding so it can be managed from View;
* [Text] Add line-height to overline (for multiline/combinations)

### DIFF
--- a/src/UI/Icon.elm
+++ b/src/UI/Icon.elm
@@ -18,12 +18,18 @@ module UI.Icon exposing
     , toEl
     , todo
     , toggle
+    , toggleDown
+    , toggleUp
+    , withColor
     )
 
 import Element exposing (..)
+import Element.Font as Font
 import Html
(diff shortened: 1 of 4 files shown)


src/UI/Icon.elm
```diff
@@ -38,8 +44,17 @@ type alias Properties =
     }
 
 
+type alias Options =
+    { color : IconColor }
+
+
+type IconColor
+    = ColorFromPalette Palette.Color
+    | ColorInherit
```
jouderianjr, line 53: I would leave the black color as default instead of having a color Inherited (same for texts btw)

PedroHLC: This can be done, but it will require another PR, fixing the color everywhere icons are used in WMS.

PR #23, opened 2020-05-25
Allow ellipsis to apply in more than just a simple one-line text.

PedroHLC:
A non-breaking, easier to comprehend version of 71be1ad

### DIFF
--- a/src/UI/Internal/Text.elm
+++ b/src/UI/Internal/Text.elm
@@ -1,6 +1,6 @@
 module UI.Internal.Text exposing (..)
 
-import Element exposing (Attribute, Element)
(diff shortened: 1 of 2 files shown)


src/UI/Internal/Text.elm
```diff
@@ -11,7 +11,7 @@ import UI.Utils.Element as Element
 
 
 type Text
-    = Text (List Span)
+    = Text (List Span) CombinedOptions
```
jouderianjr, line 14: What do you think about having `TextOptions` and `SpanOptions`?

PedroHLC: I like it!

PR #27, opened 2020-05-29
WMS-237 | Add size modifiers

PedroHLC:


PedroHLC:
image
@jouderianjr this is the why I propose it to be in RenderConfig instead of each component (and manually affecting sub-componets)

docs/src/Buttons/Stories.elm
```diff
@@ -187,3 +197,82 @@ Button.bodyIcon (Icon.toggle "Some Hint")
     |> Button.toEl renderConfig
 ```
 """ } )
+
+
+contextualSizeStory renderConfig =
+    let
+        collectionOfButtons cfg =
```
jouderianjr, line 204: So, I still think this breaks the API in some way. setting the component size in the `RenderConfig` lead us to do things like:

```elm
Alert.danger "foo bar"
    |> Alert.toEl 
         (RenderConfig.withContextualSize cfg RenderConfig.LargeSize)

``` 

```elm
Paginator.nonNumeric
    { onNextButtonClicked = NextPage
    , onPreviousButtonClicked = PreviousPage
    , totalCount = List.length paginatorsStories.items
    , offset = paginatorsStories.offset
    , first = defaultPageSize
    }
    (RenderConfig.withContextualSize cfg RenderConfig.LargeSize)
```

or even

```elm
    Nav.navigator ForNav
        navState
        (toNavContainer context >> Nav.containerMap ForPage)
        |> Nav.withMenuPages (menuPageList locale page)
        |> Nav.withMenuActions
            [ Nav.menuAction
                (Icon.logout (t "sidebar.logout"))
                SessionClear
            ]
        |> Nav.withMenuLogo (t "common.paackSlogan")
            (Vectors.paackLogoWhite (Palette.color toneGray brightnessLight))
        |> Nav.toEl 
            (RenderConfig.withContextualSize cfg RenderConfig.LargeSize)
            page
```

and what does that mean? what will be affected in the navigator if we set to `Large`?

---- 

I understand that having this in the `RenderConfig` can be useful in some use cases but I think we are making your API a bit not strict and open to inconsistent things like a mentioned. 

Also, in my point of view, the sizing modifiers are in the same "category" of the color ones. A lot of items can have different colors should we set the color in the RenderConfig?

Last but not least, what are the other components we already now that need this modifier? I think by now the only one is the button, so setting this only for one component use case maybe is not needed. if in the future we realize this could be better for some reason to RenderConfig, we do it.

Last but not least V2.0.0 I think the API is more explicit and declarative using the component modifiers, it is more easily "understandable" that this button size is being modified by that function than something inside of something else

PR #28, opened 2020-06-08
Update dialog

PedroHLC:
This PR:
* Update dialog padding to match Zeplin
* ~Use `<Button:ToneClear>` instead of `Icon`, this way padding and aria-helpers are concise.~ (REVERTED)
* Fix an width issue where `row > collumn` was rendered as `div > s > div` and the `s` element wasn't properly respecting `width:100%` rule. For this, the `row` was replaced with a simpler `el`

Required changes in parent projects:
* Remove padding in dialog's body element.

### DIFF
--- a/src/UI/Internal/Dialog.elm
+++ b/src/UI/Internal/Dialog.elm
@@ -57,10 +57,14 @@ view cfg { title, body, close, width } =
                         , Element.alignTop
                         ]
                 , Icon.close "Close dialog"
-                    |> Button.bodyIcon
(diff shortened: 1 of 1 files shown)


jouderianjr:
Screenshot 2020-06-08 at 16 40 28

the button icon is blue 🔵 , it should be black/grayish

Also, maybe I'm confused about this 8px grid, but setting the x padding to 20 is not wrong according with that rule? it should be 16 or 24px right? should we ask sufiyan?

PR #32, opened 2020-06-11
WMS-280 | Custom Icons

jouderianjr:
This is not exactly a proper PR, is more a `show and tell` then we see what we will actually implement.

So, welcome to Jouderian's struggle, please have a sit 🛋️ .

DISCLAIMER: I will be adding more explanation as comments in the code.

So, after some research I kinda found that SVG is the best way to integrate custom icons, it is lightweight and with good performance.

So, we have some steps to achieve our goal:

1. Extract the icons from a sketch file
2. Optimize icons
3. Generate an icon sprite.
4. Inject this into html.

We do have tools for all the steps (almost :o ):

1. Extract the icons from a sketch file - Sketchtool
2. Optimize icons - [svgo](https://github.com/svg/svgo)
3. Generate an svg sprite - [svg-sprite](https://github.com/jkphl/svg-sprite)
4. Inject this into HTML - We can use posthml to inject only needing to install `posthtml-include`

After that we would be able to do things like: 

```elm
    Element.html <|
        Svg.svg
            [ SvgAttrs.width "100%"
            , SvgAttrs.height "100%"
            , SvgAttrs.fill "currentColor"
            ]
            [ Svg.use [ SvgAttrs.xlinkHref ("#" ++ iconId) ] []
            ]
```

What is the problem in that flow?

Is basically in the last step. Elm takes control of the whole `<body>` so we are not able to inject code in there as elm will override. 

Then how would we inject the svg sprite there?

1. We could create a custom element and inject this custom element in the root of our Html view. (that what we have in this pr). The problem is because we generate 2 files that need to be served, `icons.html` and `svgImport.js`.

2. We could have a step that transforms all the svg code in elm and then we can ship inside of `paack-ui`. It will be quite as the first option as we will have to inject the svg in the root of our view, but the good part is that we are will not need to serve non-elm files.

So, in the end this way I did works, but how are we going to serve the js and svg file?

Edit-1: sketchtool only works on Mac :o


docs/public/icons.html
```diff
@@ -0,0 +1 @@
...97-3.145-.114-.102a1.526 1.526 0 00-1.902 0l-.115.102-.103.112a1.466 1.466 0 000 1.873l.103.112 3.196 3.147-3.196 3.147a1.467 1.467 0 000 2.098 1.525 1.525 0 002.017.101l.114-.101L16 19.503l3.197 3.147.114.101c.553.444 1.35.444 1.902 0l.115-.101.103-.113c.45-.544.45-1.329 0-1.873l-.103-.112-3.197-3.147 3.197-3.147a1.467 1.467 0 000-2.097 1.525 1.525 0 00-2.017-.102zM27.325.047A4 4 0 0131.288 3.5H.711A4.002 4.002 0 014.675.047h22.65z" fill="currentColor" fill-rule="nonzero"/></g></symbol><symbol viewBox="0 0 36 34" id="Warning1" xmlns="http://www.w3.org/2000/svg"><g fill="none" fill-rule="evenodd"><path d="M-17-18h70v70h-70z"/><path d="M21.102 1.814L35.35 26.687c1.33 2.322.556 5.3-1.728 6.652a4.728 4.728 0 01-2.409.661H4.787C2.143 34 0 31.822 0 29.135c0-.86.224-1.704.65-2.448L14.897 1.814A3.558 3.558 0 0119.806.496a3.62 3.62 0 011.296 1.318zm-4.897 8.997l.455 10.028c0 1.038.446 1.81 1.34 1.81.894 0 1.34-.772 1.34-1.81l.454-10.028c0-1.038-.803-1.88-1.794-1.88-.992 0-1.795.842-1.795 1.88zM18 30.24c1.321 0 2.393-1.123 2.393-2.507 0-1.385-1.072-2.507-2.393-2.507-1.322 0-2.394 1.122-2.394 2.507 0 1.384 1.072 2.507 2.394 2.507z" fill="currentColor" fill-rule="nonzero"/></g></symbol></svg>
```
jouderianjr, line 1: Why this is a `.html` instead of `.svg`? is because parcel understand HTML code and just reads like a string :D

PedroHLC:
> 1. Extract the icons from a sketch file
Take a look at the new Figma :smile:

docs/src/Icons.elm
```diff
@@ -33,18 +43,24 @@ iconView cfg ( iconFn, label ) =
...
     icons
         |> List.map (iconView cfg)
+        |> (::) svgSprite
         |> Element.wrappedRow
             [ Element.spacing 10
             ]
         |> Element.layout []
+
+
+svgSprite =
```
jouderianjr, line 61: This is the code we would have to inject in the root of our view

src/UI/Icon.elm
```diff
@@ -192,6 +201,7 @@ toEl cfg (Icon { hint, glyph } { color, size }) =
             , Element.centerX
             , Font.center
             , Element.width <| Element.px width
+            , Element.height <| Element.px height
             , Font.size height
```
jouderianjr, line 205: this Font.size can be removed once we remove all the fontawesome icons

docs/public/icons.html
PedroHLC, line 1: Can we generate this file instead?
Something like creating a directory "assets/icons", put unoptimized SVGs there, and then use a parcel plugin to optmize and concatenate them all?

src/UI/Icon.elm
```diff
@@ -291,16 +304,23 @@ defaultOptions =
     }
 
 
+svgIcon : String -> Element msg
+svgIcon iconId =
+    Element.html <|
+        Svg.svg
+            [ SvgAttrs.width "100%"
+            , SvgAttrs.height "100%"
```
PedroHLC, line 312: Can we adopt that all icons will be 1:1, and replace `height:100%` with preversing aspect-ratio?

PedroHLC:
> sketchtool only works on Mac :o

STALLMAN TRIGGERED

LINUS FUCK

PedroHLC:
I think you should replace those nested npm commands with a Makefile :sweat_smile:

PedroHLC:
Also what happened to the last two icons?

jouderianjr:
Yes! it's working, ready to go @PedroHLC review again

jouderianjr:
Btw, this pr doesn't try to solve all the icon sizes. I prefer to do it in another one

PR #34, opened 2020-06-15
WMS-289 | Adjust to Figma's parameters

PedroHLC:
* Rename ContextualSize => Size
* Update Size types/values
* Update Icon sizes
* Update Button colors, borders, paddings
* Update Text's height
* Update Palette.gray.light

Note: This will break WMS, because of new Size values.


PedroHLC:
> In the sizes notes, I would add something like:
> 
> """
> import UI.Size as Size
> 
> Size.smalSize
> Size.smalMedium
> Size.smalLarge
> """

Good catch!

PR #35, opened 2020-06-15
WMS-292 | Add ToggableList as Table's mobile

PedroHLC:


PedroHLC:
Deployed to WMS's sandbox to easy previewing.

src/UI/Internal/ToggableList.elm
```diff
@@ -0,0 +1,123 @@
...
+        , Element.height Element.shrink
+        ]
+        [ config.coverView renderConfig (titleColor selected) object selected
+            |> Element.el
+                [ Element.width fill
+                , Element.centerY
+                , Element.clipX
+                , Element.paddingXY 0 5
+                ]
+        , ifThenElse selected
```
jouderianjr, line 57: That's nice usage :o never thought about it 🕺

src/UI/Internal/ToggableList.elm
```diff
@@ -0,0 +1,123 @@
...
+    Element.column [ Element.width fill ]
+        [ defaultRow renderConfig config True object
+        , object
+            |> config.details
+            |> List.map (someCard renderConfig)
+            |> Element.column toggleableCard
+        ]
+
+
+someCard : RenderConfig -> ( String, Element msg ) -> Element msg
```
jouderianjr, line 83: I'm not sure about this name, maybe we can have something more explicit? like fieldCard? mobileFieldCard? or something better than my poor suggestions hahah

src/UI/Table.elm
```diff
@@ -26,27 +28,45 @@ import UI.Button as Button exposing (Button)
...
 
-type alias Options msg columns =
-    { rows : List (Row msg columns)
+type alias Options msg object columns =
+    { rows : Rows msg object columns
     , width : Element.Length
     }
 
 
+type alias ResponsiveOptions msg object columns =
```
jouderianjr, line 54: This case I'm not sure if we should call ResponsiveConfig or Options 🤔 , I think in the pr that I added this idea of having a record for more than `x` fields in a `with*` function I called Config

src/UI/Table.elm
```diff
@@ -26,27 +28,45 @@ import UI.Button as Button exposing (Button)
...
+    { detailsShowLabel : String
+    , detailsCollapseLabel : String
+    , selectMsg : object -> msg
+    , isSelected : object -> Bool
+    , coverView : RenderConfig -> Palette.Color -> object -> Bool -> Element msg
+    , toRow : object -> Row msg columns
```
jouderianjr, line 60: Ok so thaaat's is the complicated part, if we look at the Events page, we have a duplicated info:

Screenshot 2020-06-17 at 13 12 35

I think in Zeplin the `created at` does not appear in the details, is something like, we would need to choose what fields we want to show in the detailed version 🤔

docs/src/Tables.elm
```diff
@@ -1,12 +1,18 @@
...
 import UI.Table as Table
     exposing
-        ( cellWidthEnd
+        ( cellFromText
+        , cellWidthEnd
         , cellWidthPortion
         , cellWidthShrink
+        , detailsEnd
+        , detailsHide
+        , detailsShow
```
jouderianjr, line 15: Maybe `mobileDetailsShow/Hide/End`, as this module exposes a lot of things, having this prefix makes it more self explanatory

src/UI/Internal/NList.elm
```diff
@@ -36,6 +36,11 @@ map2 applier (NList xs) (NList ys) =
     NList <| List.map2 applier xs ys
 
 
+map3 : (a -> b -> c -> result) -> NList a n -> NList b n -> NList c n -> NList result n
```
jouderianjr, line 39: is it being used? if not should we remove? 🤔

src/UI/Internal/ToggableList.elm
```diff
@@ -0,0 +1,161 @@
...
+import UI.Size as Size
+import UI.Text as Text
+import UI.Utils.Element exposing (zeroPadding)
+
+
+type alias Config object msg =
+    { detailsShowLabel : String
+    , detailsCollapseLabel : String
+    , toCover : object -> Cover
+    , toDetails : object -> List ( String, Element msg )
```
jouderianjr, line 21: nice, it is better than `toRow`

PR #37, opened 2020-06-18
WMS-306 | API Perfumery

PedroHLC:
This PR includes some API-breaking modificatios:
* Rename `*.toEl` to `*.renderElement`
* Rename `Palette.toElColor` to `Palette.toElementColor`
* Rename `Link.packEl` to `Link.wrapElement`
* Button API.v2

After this, we can move to documentation!

### DIFF
--- a/docs/src/Buttons/Stories.elm
+++ b/docs/src/Buttons/Stories.elm
@@ -10,7 +10,7 @@ import UI.Icon as Icon
 import UI.Link as Link
 import UI.RenderConfig as RenderConfig
 import UIExplorer exposing (storiesOf)
-import Utils exposing (story, storyList, storyWithModel)
+import Utils exposing (iconsSvgSprite, story, storyList, storyWithModel)
 
 
 update : Buttons.Msg -> Buttons.Model -> Return Buttons.Msg Buttons.Model
@@ -38,7 +38,8 @@ stories renderConfig =
 enabledStory renderConfig label tone toneStr =
     storyList
         ( label
-        , [ Button.fromLabel "Prompt"
+        , [ iconsSvgSprite
+          , Button.fromLabel "Prompt"
                 |> Button.cmd Msg.NoOp tone
                 |> Button.renderElement renderConfig
           , Button.fromIcon (Icon.toggle "Toggle")
@@ -67,7 +68,8 @@ Button.fromIcon Icon.someIcon
 disabledButtonStory renderConfig =
     storyList
         ( "Disabled"
-        , [ Button.fromLabel "Prompt"
+        , [ iconsSvgSprite
+          , Button.fromLabel "Prompt"
                 |> Button.disabled
                 |> Button.renderElement renderConfig
           , Button.fromIcon (Icon.toggle "Toggle")
@@ -112,7 +114,8 @@ dangerStory cfg =
 successStory renderConfig =
     storyList
         ( "Success"
-        , [ Button.fromLabel "Prompt"
+        , [ iconsSvgSprite
+          , Button.fromLabel "Prompt"
                 |> Button.success
                 |> Button.renderElement renderConfig
           , Button.fromIcon (Icon.toggle "Toggle")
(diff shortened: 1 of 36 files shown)


src/UI/Button.elm
```diff
@@ -32,283 +31,502 @@ import Element.Border as Border
...
+type ButtonAction msg
+    = ActionMsg msg
+    | ActionRedirect Link
 
 
-type ButtonMode
-    = ModeEnabled
-    | ModeDisabled
+type ButtonMode msg
+    = ButtonWorking (ButtonAction msg) ButtonStyle
```
jouderianjr, line 75: Maybe `ButtonActive`? Because in my opinion,`ButtonWorking` seems that the button is doing something in that particular moment, but in fact, it is like waiting for some external action, like `idle`

src/UI/Button.elm
```diff
@@ -32,283 +31,502 @@ import Element.Border as Border
...
 
-link : Link -> ButtonBody -> Button msg
-link realLink =
-    buttonAny (ClickLink realLink)
 
+-- Body to Button
 
 
--- Options
+disabled : ButtonBody -> Button msg
```
jouderianjr, line 134: So, maybe we don't need to have disabled and success as creation functions, I think those 2 states are always conditionally linked to something. For example. I want to disable the button when the text field is empty. I don't see a case that someone is using those functions instead of `withDisableIf` or `withSuccessIf`.

src/UI/Button.elm
```diff
@@ -32,283 +31,502 @@ import Element.Border as Border
...
+fromIcon : Icon -> ButtonBody
+fromIcon icon =
+    BodyIcon icon
 
 
-link : Link -> ButtonBody -> Button msg
-link realLink =
-    buttonAny (ClickLink realLink)
 
+-- Body to Button
```
jouderianjr, line 131: I will call this type of functions in a different way. Maybe entrypoints, create/init functions or something like that as this concept can be applied across all our components, would be nice to have one term to that type of function

src/UI/Button.elm
```diff
@@ -340,269 +578,192 @@ type alias ButtonTheme =
...
-            else
-                ( Palette.toElColor primary
-                , Palette.toElColor primary
-                , text
-                    |> Text.fontColor
-                    |> Maybe.withDefault
-                        Palette.gray.middle
-                )
-    in
+themeApply : ButtonTheme -> List (Attribute msg)
```
jouderianjr, line 581: Looking at the type signature, I would say that this function should be called `themeToAttributes ` and not the other one.

docs/src/Buttons/Stories.elm
```diff
@@ -104,49 +100,74 @@ Button.bodyIcon Icon.someIcon
...
 
 
 linkStory renderConfig =
     story
         ( "Link"
-        , Button.bodyText "Go to Blank"
-            |> Button.link (Link.link "about:blank")
-            |> Button.toEl renderConfig
+        , Button.fromLabel "Go to Blank"
+            |> Button.redirect (Link.link "about:blank") Button.hyperlink
```
jouderianjr, line 164: Yeah, I think we don't need Link module anymore, this line could be only: 

`|> Button.redirect "about:blank" Button.hyperlink`

and then we move everything is needed from UI.Link to this module as internal functions

src/UI/Button.elm
PedroHLC: I found this hard: We have only one simple init "toggle" that doesn't need a body cause it's always the same: Then, we have some action-based init "cmd", and "redirect", and some static init "disabled" and "success". These last three require a body.

It's hard to fit them all in one term when what they do happens in different ways.

Also "init" is a term for system initializations, like `SDL_init(..)`, `OGLinit(..)`, not data that later may be duplicated and became four-five buttons in a list. (Considering this duplication we could also discard "new" and "create")

Other options I could think of are "define", "assemble", "begin".

src/UI/Button.elm
PedroHLC: That's what the function below is called :eyes: 

I'll do something anyway...

docs/src/Buttons/Stories.elm
jouderianjr: if is that the case, we can move to an Internal module and still do something like:

`|> Button.redirect "about:blank" Button.hyperlink`

I'm not sure where the user would use this link, if only the components are using, we can set as an internal thing

src/UI/Button.elm
jouderianjr: but it's not about requiring one value or another, is about being the start point of that builder.

In that cases you mentioned, all of them are the entry point, it means that they receive a bunch of things and return a Button, then that's different of the modifiers (with* functions). So in that case, `toggle`, `cmd` and `redirect` are from the same category, they are the functions used to get a type `Button`

PedroHLC: Done

PR #39, opened 2020-06-19
WMS-307 | (Pt 1) Add LICENSE, finish JSONs and remake Showcase

PedroHLC:


elm.json
```diff
@@ -1,7 +1,7 @@
 {
     "type": "package",
     "name": "PaackEng/paack-ui",
-    "summary": "Paack design system",
+    "summary": "Paack's Design System applied in components for Elm UI",
```
jouderianjr, line 4: I think by definition, Design system already implies that is in components 🤔 , so maybe "Paack's Design System for Elm UI"

package.json
```diff
@@ -1,29 +1,36 @@
...
     "doc": "docs"
   },
   "scripts": {
-    "watch-docs-old": "chokidar \"**/*.elm\" -c \"elm make docs/src/Main.elm\"",
-    "watch-docs": "parcel docs/public/index.html",
-    "build-docs": "parcel build docs/public/index.html",
-    "test": "elm-test",
-    "analyse": "elm-analyse"
+    "analyse": "elm-analyse",
+    "docs": "elm make --docs=dist/docs.json",
```
jouderianjr, line 11: why use elm make instead of parcel? this way we lose the local server and the hot reload.

PedroHLC: That is for previewing the docs, not related to the showcase

PR #41, opened 2020-06-22
WMS-307 | Button's elm-compatible documentation

PedroHLC:
The first of many to come.

### DIFF
--- a/src/UI/Button.elm
+++ b/src/UI/Button.elm
@@ -1,29 +1,72 @@
 module UI.Button exposing
-    ( Button
-    , ButtonBody
-    , ButtonStyle
-    , ButtonWidth
-    , clear
-    , cmd
-    , danger
-    , disabled
-    , fromIcon
-    , fromLabel
-    , full
-    , hyperlink
-    , light
-    , map
-    , primary
-    , redirect
-    , renderElement
-    , shrink
-    , success
-    , toggle
-    , withDisabledIf
+    ( Button, toggle, success, disabled, cmd, redirect
+    , ButtonBody, fromLabel, fromIcon
+    , ButtonStyle, hyperlink, primary, danger, light, clear
+    , ButtonWidth, withWidth, full, shrink
     , withSize
-    , withSuccessIf
-    , withWidth
+    , withDisabledIf, withSuccessIf
+    , renderElement
+    , map
     )
+
+{-| The button is the unified aggregation of all the standardized clickable elements.
+The same component can render as a hyperlink, an embossed button, or a padded-icon.
+
+The embossed-styled button has four tones: Primary, Danger, Light, Clear.
+And it can also be in one of the states: Active, Disabled, and Success.
+
+Following Elm-UI standards, this component is accessible.
+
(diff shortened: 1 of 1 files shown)


src/UI/Button.elm
```diff
@@ -205,6 +346,15 @@ withWidth width button =
             Button prop { opt | width = width }
 
 
+{-| With `Button.withSize`, you'll be able to scale the button between the [standard sizes][size].
+
+[size]: UI-Size
+
+    Button.withSize Size.large someButton
+
+**NOTE**: Default value is [`Button.shrink`](UI-Button#shrink)
```
jouderianjr, line 355: This was copied from the other function, but I think it's `Size.large` right? 🤔

src/UI/Button.elm
```diff
@@ -216,48 +366,69 @@ withSize size button =
...
+{-| The button's width will fill its container
+-}
 full : ButtonWidth
 full =
     WidthFull
 
 
+{-| The button will have the exact width to fit its contents.
+-}
 shrink : ButtonWidth
```
jouderianjr, line 421: Maybe add a Note pointing that this one is the default.

src/UI/Button.elm
```diff
@@ -1,30 +1,73 @@
...
+    , withDisabledIf, withSuccessIf
+    , renderElement
+    , map
     )
 
+{-| The button is the unified aggregation of all the standardized clickable elements.
+The same component can render as a hyperlink, an embossed button, or a padded-icon.
+
+The embossed-styled button has four tones: Primary, Danger, Light, Clear.
+And it can also be in one of the states: Active, Disabled, and Success.
```
jouderianjr, line 16: Non-pr related, maybe we should call `Enabled` instead of `Active` as we are using `Disabled` and not `Deactivated`

src/UI/Button.elm
```diff
@@ -1,30 +1,73 @@
...
+    , ButtonWidth, withWidth, full, shrink
     , withSize
-    , withSuccessIf
-    , withWidth
+    , withDisabledIf, withSuccessIf
+    , renderElement
+    , map
     )
 
+{-| The button is the unified aggregation of all the standardized clickable elements.
```
jouderianjr, line 12: Maybe I don't need to expose that we treat the link and embossed button are different way. we could only say:

"The same component can render as a hyperlink, primary, danger, light and clear"

src/UI/Button.elm
jouderianjr: @PedroHLC I think  you changed in the wrong place hahaha

src/UI/Button.elm
```diff
@@ -195,6 +325,13 @@ withDisabledIf condition button =
         button
 
 
+{-| With `Button.withWidth`, you'll be able to set one of the available values for width.
+
+    Button.withWidth Button.fill someButton
+
+**NOTE**: Default value is [`Size.medium`](UI-Size#medium)
```
jouderianjr, line 332: is the opposite right?

PR #45, opened 2020-06-26
WMS-307 | Insert docummentation to all exposed components

PedroHLC:
Included changes not related:
* Renamed SelectList to RowList;
* Renamed Toggable to Toggleable;
* Bunch of cleanups.


PedroHLC:
For previewing, access:
https://elm-doc-preview.netlify.app/?repo=PaackEng/paack-ui&version=v1-docs-comments-accumulated
& Submit:
https://gist.githubusercontent.com/PedroHLC/ec4bf63534f5da4f116a745fee59766b/raw/paack-ui.docs.json

src/UI/Button.elm
```diff
@@ -144,7 +144,7 @@ type ButtonStyle
 
 {-| To maintain fewer variations of similar-looking content, we opted to have only two sizes of buttons:
 One with the minimum to fit its contents, and the other filling the container.
-The `ButtonWidth` type describes these two values.
+The `ButtonWidth` type work with these two values.
```
jouderianjr, line 147: Why did you change this? I think describes makes more sense :D

src/UI/Button.elm
```diff
@@ -325,11 +325,11 @@ withDisabledIf condition button =
         button
 
 
-{-| With `Button.withWidth`, you'll be able to set one of the available values for width.
+{-| `Button.withWidth` changes the width of the button to either have the minimum required for its contents or fill the parent horizontally.
```
jouderianjr, line 328: I think it's `required by its parent`, but I'm not sure tho 🤔

src/UI/Button.elm
```diff
@@ -505,12 +505,12 @@ toggleView cfg size hint toggleMsg current =
...
 
         attrs =
-            [ ARIA.roleAttr ARIA.roleButton
-            , Primitives.roundedBorders
+            [ Primitives.roundedBorders
             , paddings
             , borders
             , Events.onClick <| toggleMsg (not current)
             ]
+                ++ ARIA.toElementAttributes ARIA.roleButton
```
jouderianjr, line 513: Maybe it makes sense only to have a function called ARIA.roleButtonAttributes returning the list of Attributes? why do we need to have this type of function (`toElementAttributes`)?

src/UI/Internal/ToggleableList.elm
```diff
@@ -1,4 +1,4 @@
-module UI.Internal.ToggableList exposing (Config, Cover, view)
+module UI.Internal.ToggleableList exposing (Config, Cover, defaultRow, selectedRow)
```
jouderianjr, line 1: Do we need to expose `defaultRow` and `selectedRow`?

src/UI/Link.elm
```diff
@@ -20,30 +37,57 @@ type alias Properties =
     { href : String }
 
 
+{-| The `Button msg` type is used for describing the component for later rendering.
```
jouderianjr, line 40: `Link msg`

src/UI/NavigationContainer.elm
```diff
@@ -1,28 +1,86 @@
...
+                    Msg.SessionLogout
+                ]
+            |> Nav.withMenuLogo "My company's logo" someLogoElement
+            |> Nav.toBrowserDocument renderConfig currentPage
+
+    getPageContainer : Page.Page -> Nav.Container Page.Msg
+    getPageContainer page =
+        case page of
+            Page.Packages ->
+                { title = "Pacakges"
```
jouderianjr, line 41: "Packages"

src/UI/NavigationContainer.elm
```diff
@@ -39,40 +97,70 @@ import UI.Utils.Element as Element
 -- Expose
 
 
+{-| This record must be generated with [`Nav.menuPage`](UI-NavigationContainer#menuPage)
```
jouderianjr, line 100: Ok, so maybe if we want to make it accessible only using `Nav.menuPage` we must make this an opaque type ( I think it's a good idea).

Same for `MenuAction`

src/UI/NavigationContainer.elm
```diff
@@ -39,40 +97,70 @@ import UI.Utils.Element as Element
...
 
 
+{-| Keep this one in your Model, it holds the current navigation state.
+-}
 type alias State =
-    -- Keep this one in MODEL
     { menuExpanded : Bool }
 
 
+{-| The `Nav.Msg` handles menu related messages transparently.
```
jouderianjr, line 122: What do you want to say by "transparently"?

src/UI/NavigationContainer.elm
```diff
@@ -168,31 +318,92 @@ stateUpdate msg state =
 -- Constructors
 
 
+{-| The correct way of instantiating a [`Nav.State`](UI-NavigationContainer#State).
```
jouderianjr, line 321: I think `Nav.State` must be opaque too, this way, if we add more features, it is less probably to be a breaking change.

src/UI/NavigationContainer.elm
```diff
@@ -168,31 +318,92 @@ stateUpdate msg state =
...
+
+-}
 menuAction : Icon -> msg -> MenuAction msg
 menuAction icon msg =
     Menu.Action icon msg
 
 
+{-| `Nav.navigator` holds the minimum amount of information required for all the features (menu, dialogs, and page's layout) to work.
+
+The first parameter is a lambda that should transform [`Nav.Msg`](UI-NavigationContainer#Msg) in a message type controlled by the user.
```
jouderianjr, line 389: I don't think `lambda` is a very used term in the Elm community, maybe I would change to function

src/UI/NavigationContainer.elm
```diff
@@ -215,12 +438,20 @@ dialog title onClose body =
...
+    Nav.toBrowserDocument renderConfig currentPage navigator
+
+-}
+toBrowserDocument :
     RenderConfig
     -> page
     -> Navigator page msg
     -> { body : List (Html msg), title : String }
-renderElement cfg page model =
+toBrowserDocument cfg page model =
```
jouderianjr, line 454: 👍

src/UI/RenderConfig.elm
```diff
@@ -57,22 +95,49 @@ updateWindow window (RenderConfig oldData) =
...
+    else
+        -- Desktop view
+
+-}
 isMobile : RenderConfig -> Bool
 isMobile (RenderConfig { deviceClass }) =
     deviceClass == Element.Phone
 
 
+{-| `True` when the browser is in portrait size and not landscape.
```
jouderianjr, line 111: Not correct text

src/UI/RenderConfig.elm
```diff
@@ -57,22 +95,49 @@ updateWindow window (RenderConfig oldData) =
...
 elLayoutAttributes : RenderConfig -> List (Attribute msg)
 elLayoutAttributes _ =
     -- Why here? Accessibility settings may change fonts, backgrounds, etc...
     [ Font.family [ Font.typeface "Inter", Font.sansSerif ] ]
 
 
+
+-- Internal
+
+
```
jouderianjr, line 140: Maybe we don't need that 🤔  I'm not sure we are using this as a pattern in all other modules

src/UI/RowList.elm
```diff
@@ -0,0 +1,469 @@
...
+    , withOptions, withSelected
+    , SearchConfig, withSearchField, ActionConfig, withActionBar
+    , withWidth
+    , renderElement
+    )
+
+{-| `UI.RowList` is a styled searchable row list.
+The main variation (select-lists) has the capability of having one of its rows selected.
+
+The developer does the rendering of the row elements.
```
jouderianjr, line 13: I think this phrase is a bit confusing, maybe it would be better to say that the developer is in charge of creating the "Row" view?

src/UI/RowList.elm
```diff
@@ -0,0 +1,469 @@
+module UI.RowList exposing
```
jouderianjr, line 1: So, I'm not very sure about this name 🤔  I think some frameworks would call this ListView. React-native that calls `FlatList` and I kinda like this name, WDYT?

jouderianjr: I just realize that the lists are merged here, so maybe just call ListView would be better, it is more "generic", not sure tho, up to you

src/UI/RowList.elm
```diff
@@ -0,0 +1,469 @@
...
+{-| Replaces the component's list of elements.
+
+    RowList.withOptions
+        [ { id = 0, name = "Catarina" }
+        , { id = 1, name = "Gabriel" }
+        ]
+        someRowList
+
+-}
+withOptions : List object -> RowList object msg -> RowList object msg
```
jouderianjr, line 275: I know it's not exactly related to this PR, but I don't think `withOptions` is a good name, we use the word `Options` internally a lot of another thing, so maybe it can be confusing. I would suggest `withItems` or `withData`

src/UI/Text.elm
```diff
@@ -1,110 +1,203 @@
...
+    , heading1, heading2, heading3, heading4, heading5, heading6
+    , subtitle1, subtitle2
+    , caption, overline
+    , multiline, combination
     , withColor
+    , setEllipsis
+    , renderElement
     )
 
+{-| `UI.Text` is a component to specify how text to display text. It applies font size, weight, letter-spacing, and color.
```
jouderianjr, line 13: I would enforce in the description that we should not be using Element.text directly in paack's project 🤔

src/UI/Button.elm
PedroHLC: There are some ARIA attributes like "aria-describedby" that doesn't need to be attached to a role, the idea is that we may put those in the future using with-pattern.

src/UI/Internal/ToggleableList.elm
PedroHLC: We can't import the main list without creating a recursive import. So what leaves is to let this file atomic and keep it internal, as I did here.

src/UI/RenderConfig.elm
PedroHLC: In a few, it doesn't interfere in docs.

jouderianjr: Yeah, but let's try to be consistent then, add this everywhere

PR #47, opened 2020-06-30
WMS-315 | Phase 3 filtered-sortable table (pt 1)

PedroHLC:
[Table] Add filtering
[Primitives] Add relatively-rounded borders and remove rounded fields
[Palette] Add overlayBackground and mainBackground
[Table] some cells are better-named columns
[Button] Add nested body

### DIFF
--- a/showcase/src/Tables/Model.elm
+++ b/showcase/src/Tables/Model.elm
@@ -1,12 +1,47 @@
-module Tables.Model exposing (Model, initModel)
+module Tables.Model exposing (Column(..), Model, ModelFilters, getColumnFilter, initModel)
 
 
 type alias Model =
     { selected : Maybe String
+    , filters : ModelFilters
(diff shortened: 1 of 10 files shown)


PedroHLC:
Currently looks like:
Screenshot

PedroHLC:
To make reviewing easy, I'm splitting this PR into two, this one creates this component:
table

The second one will:
  * Resize textfield;
  * Use the proper icons;

PedroHLC:
> The "popup" is appearing below the table header position, it should be on top
It's not, at least works on my machine> But, I will investigate.

>  It seems that should be only a clickable icon
It has a hovering state according to Figma, but you're right, I'll change to only the icon be clickable.

PedroHLC:
Closing in favor of a new API aggregating better all use-cases.

PR #50, opened 2020-07-02
WMS-315 | Phase 3 filtered-sortable table (Table v3) - part 1

PedroHLC:
Other included changes:
```
[Palette][Internal] Add mainBackgrond and overlayBackground

[Button] Add fromNested and fromAtomicNest; Use relativeRoundedBorders
    -- Still requires the behavior to click on Icon only

[NList] Add NArray

[Primitives][Internal] Add relativeRoundedBorders
```


PedroHLC:
I again divided it into two PRs, otherwise, it'll be too big for reviewing!

This is how it looks like now:
Demonstration

PedroHLC:
@jouderianjrm you're gonna notice that the applied filter is visually wrong, that the buttons still don't close the dialog, and that the other filters visuals aren't read yet. You're not supposed to be reviewing those in this PR, leave it to the late one.

On this PR please review: new Table and Filters API, functions and variables nomenclatures, record/units formats, nested button's API.

src/UI/Table.elm
```diff
@@ -1,64 +1,121 @@
 module UI.Table exposing
-    ( Table, table
-    , HeaderRow, headersEnd, header
-    , Row, withStaticRows, rowEnd, cellFromText, cellFromButton
-    , OptRow
-    , CellWidth, withWidth, withCellsWidth, cellWidthEnd, cellWidthPixels, cellWidthPortion
-    , withResponsiveRows, ResponsiveConfig, MobileCover
-    , withCellsDetails, cellMobileDetailsEnd, cellMobileDetailsHide, cellMobileDetailsShow, cellMobileDetailsShowIf
+    ( Table, table, withItems
+    , Columns, columnsEmpty, columnsPushHeader, columnsPush
```
jouderianjr, line 3: I'm not sure about the `Push` term here (in the module in general).

here I would prefer `column`

```elm
            columnsEmpty
                |> column (headerToColumn "Title" |> columnWidthPixels 320)
                |> column (headerToColumn "Author" |> columnWidthPixels 240)
                |> column (headerToColumn "Year" |> columnWidthPixels 120)
```

Also, `columnsPushHeader` is not being used right?

jouderianjr: In fact, maybe I would change this API as well.

it could be something like: 

`column: String -> ColumnWidth -> Column`, I'm not sure if we will have any other parameter for the column 🤔

src/UI/Table.elm
```diff
@@ -1,64 +1,121 @@
...
+    , Column, headerToColumn, columnWidthPortion, columnWidthPixels
+    , Row, rowEmpty, rowPushText, rowPushButton
```
jouderianjr, line 5: Again, about the `Push` keyword. Maybe something like that?

```elm
        toTableRow { author, title, year } =
            rowEmpty
                |> rowCellText (Text.body1 title)
                |> rowCellText (Text.body2 author)
                |> rowCellText (Text.caption year)
```

showcase/src/Tables/Stories.elm
```diff
@@ -3,218 +3,103 @@ module Tables.Stories exposing (stories, update)
...
+                |> detailsPush { label = "Author", content = cellFromText <| Text.body2 author }
+                |> detailsPushHidden
+
+        toTableCover { title, year } =
+            { title = title, caption = Just year }
+
+        someFilters =
+            filtersEmpty
+                |> filtersPushSingleText Nothing (filterLocal (\{ title } str -> String.contains str title))
+                |> filtersPushSingleText (Just "Dan") (filterLocal (\{ author } str -> String.contains str author))
```
jouderianjr, line 75: So, why not implement this function inside of the Table? why this local filter would be different?

Also, this function is case sensitive, I think it should be case insensitive

src/UI/Table.elm
```diff
@@ -1,64 +1,121 @@
...
+    , Responsive, Cover, Details, Detail, withResponsive, detailsEmpty, detailsPush, detailsPushHidden
+    , Cell, cellFromText, cellFromButton
+    , State, Msg, withState, stateInit, stateUpdate
+    , Filters, withFilters, filtersEmpty, filtersPushSingleText
```
jouderianjr, line 9: Instead `filtersPushSingleText`, we could do:

```elm
        someFilters =
            filtersEmpty
                |> filter localSingleTextFilter Nothing .title
                |> filter localSingleTextFilter (Just "Dan") .author
                |> filter localSingleTextFilter Nothing .year
```

src/UI/Button.elm
```diff
@@ -266,11 +268,23 @@ fromLabel label =
...
 fromIcon icon =
     BodyIcon icon
 
 
+fromNested : String -> (String -> Icon) -> ButtonBody msg
+fromNested label icon =
+    BodyNested label icon
+
+
+fromAtomicNest : String -> (String -> Icon) -> msg -> ButtonBody msg
```
jouderianjr, line 281: I didn't like this naming, but why we would need this? I think you added this because you wanted to transform the header in button right? But I don't think it is a button, is a specific clickable view, but is not a button

src/UI/Internal/Primitives.elm
```diff
@@ -1,14 +1,29 @@
...
 roundedBorders =
     Border.rounded 8
 
 
-roundedFields : Attribute msg
-roundedFields =
-    Border.rounded 8
+{-| REVIEW: I don't think I liked this name, gimme suggestions for this!
+-}
+relativeRoundedBorders : Size -> Attribute msg
```
jouderianjr, line 15: - `roundedBordersFrom : Size -> Attribute msg`
- `roundedBorders : Size -> Attribute msg` and call the other one as `defaultRoundedBorders`

Not sure tho 🤔

src/UI/Tables/Stateable.elm
```diff
@@ -0,0 +1,746 @@
...
+{-| The `StateableTable msg item columns` type is used for describing the component for later rendering.
+-}
+type StateableTable msg item columns
+    = Table (StateableConfig msg item columns) (Options msg item columns)
+
+
+type alias StateableConfig msg item columns =
+    { columns : Columns columns
+    , toRow : ToRow msg item columns
+    , toExtern : Msg -> msg
```
jouderianjr, line 139: `toExternalMsg` maybe?

src/UI/Tables/Stateable.elm
```diff
@@ -0,0 +1,746 @@
...
+filtersPush filter accu =
+    NArray.push filter accu
+
+
+localSingleTextFilter :
+    Maybe String
+    -> (item -> String)
+    -> Filters msg item columns
+    -> Filters msg item (T.Increase columns)
+localSingleTextFilter initValue get accu =
```
jouderianjr, line 296: What do you think about moving all this logic to Filters module? I think Table should not contain the filter logic (for example, the applier), maybe we should just call a Filters function here

src/UI/Tables/Stateable.elm
```diff
@@ -0,0 +1,746 @@
...
+filterLocal isKept =
+    Filters.Local isKept
+
+
+remoteSingleTextFilter :
+    Maybe String
+    -> (String -> msg)
+    -> Filters msg item columns
+    -> Filters msg item (T.Increase columns)
+remoteSingleTextFilter initValue editMsg accu =
```
jouderianjr, line 318: Same thing

src/UI/Tables/Stateable.elm
```diff
@@ -0,0 +1,746 @@
...
+
+mobileView :
+    RenderConfig
+    -> StateableConfig msg item columns
+    -> Options msg item columns
+    -> Responsive msg item columns
+    -> Element msg
+mobileView renderConfig prop opt responsive =
+    ListView.toggleableList
+        { detailsShowLabel = "Expand"
```
jouderianjr, line 382: Not related with this pr, but having these Strings here is a problem regarding i18n

src/UI/Button.elm
```diff
@@ -501,10 +502,10 @@ toggleView cfg size hint toggleMsg current =
             iconLayout size
 
         attrs =
-            [ Primitives.roundedBorders
+            [ Primitives.roundedBorders size
             , paddings
             , borders
-            , Events.onClick <| toggleMsg (not current)
+            , Element.onIndividualClick <| toggleMsg (not current)
```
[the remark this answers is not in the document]
PedroHLC: Otherwise clicking on an element over another element triggers their both onClick.
Like the header filter view: Clicking it edits the filter while clicking on the small button over it should only remove the filter.

src/UI/Internal/Basics.elm
```diff
@@ -47,3 +47,18 @@ maybeToList maybeSomething =
...
+    case primary of
+        Just sth ->
+            Just sth
+
+        Nothing ->
+            replacement
+
+
+swap : (a -> b -> c) -> b -> a -> c
+swap applier b a =
```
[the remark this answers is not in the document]
PedroHLC: :rofl: I could swear this used to be called swap, changing it...

src/UI/Tables/Stateable.elm
PedroHLC: That used to be impossible before the "Common" file, now I can do it! Good suggestion!

src/UI/Tables/Stateable.elm
```diff
@@ -0,0 +1,746 @@
...
+
+        filters =
+            opt.filters
+                |> Maybe.map (NArray.map Just >> NArray.toList)
+                |> Maybe.withDefault (List.repeat len Nothing)
+
+        filtersState =
+            case prop.state of
+                State state ->
+                    List.map
```
jouderianjr, line 435: You could use List.indexedMap fn opt.filters this way I think `len` and `indexedList` is not required.

PR #51, opened 2020-07-08
WMS-315 | Table Filters

PedroHLC:


PedroHLC:
@jouderianjr please review the new API introduced by ac44d98.

I may be overdoing it, but it looks like this:

* Model:
```elm
    model =
       checkBoxesEmpty
           |> checkBox False
           |> checkBox False
           |> checkBox False
           |> fromCheckBoxes
```

* View:
```elm
    view cfg model =
       toView Msg.ForBoxes labels model
            |> renderElement cfg

     labels =
        labelsEmpty
           |> label "Guaraná"
           |> label "Coca-cola"
           |> label "Fanta"
```

(There is also radio behavior)

Also, I don't now if I call it `checkbox` or `checkBox`

src/UI/Internal/Human.elm
```diff
@@ -0,0 +1,32 @@
+module UI.Internal.Human exposing (parseDate)
+
+
+parseDate : String -> Maybe ( Int, Int, Int )
```
jouderianjr, line 4: I think this is not safe enough. I would use this [package](https://package.elm-lang.org/packages/rtfeldman/elm-iso8601-date-strings/latest/) to do the conversion

PedroHLC:
@jouderianjr 
I would only drop that `parseDate` for a safe, localized, and graphical time-picker. Otherwise, the user will be typing DD/MM/YYYY as proposed in Figma and not ISO's "YYYY-MM-DD HH:mm:SS.ms" :sweat_smile:

PedroHLC:
This is how it looks right now:
Demonstration

src/UI/Checkbox.elm
```diff
@@ -0,0 +1,46 @@
+module UI.Checkbox exposing (checkbox, radioButton)
```
jouderianjr, line 1: I would split into 2 modules as a radiobutton is not a checkbox.

src/UI/Checkbox.elm
```diff
@@ -0,0 +1,46 @@
...
+
+{-|
+
+
+# Atomic Elements
+
+@docs checkbox, radioButton
+
+**NOTE**:
+We have plans for "checklists" and "radioGroups", but this won't be created until usecase appears.
```
jouderianjr, line 11: So, I think we cannot have radio buttons without having the radiogroup behavior 🤔  not having the radio group now means that every time we need to use radio buttons we will need to implement the radiogroup behavior as checkboxes can show up alone, but this does not apply to radio buttons.

PedroHLC:
Non API-related changes will continue in another PR...

src/UI/Internal/FiltersHeaders.elm
```diff
@@ -287,6 +306,172 @@ singleTextFilterRender renderConfig applyMsg { fromFiltersMsg, index, label } ed
...
+            |> TextField.renderElement renderConfig
+        , radioGroup renderConfig editComparisonMsg (Just current.comparison) options
+        ]
+
+
+
+-- Internal
+
+
+dateInput : msg -> (String -> msg) -> String -> String -> DateInput -> TextField msg
```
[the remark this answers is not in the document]
PedroHLC: Yeah, with exception of `withSize, withWIdth, withOnEnterPresed`, I'll move it in #53

PR #53, opened 2020-07-13
WMS-315 | Update Checkbox and RadioButton to match Figma

PedroHLC:
[ Requires #51 ]
Also fixes an alignment thing on Table, and put hover style on filter header. (sorry)

### DIFF
--- a/showcase/src/Checkboxes/Stories.elm
+++ b/showcase/src/Checkboxes/Stories.elm
@@ -1,0 +1,100 @@
+module Checkboxes.Stories exposing (stories, update)
+
+import Checkboxes.Model as Checkboxes
+import Checkboxes.Msg as Checkboxes
+import Element
+import Msg
+import Return as R exposing (Return)
+import UI.Checkbox as Checkbox
+import UI.Internal.Basics exposing (ifThenElse)
+import UI.Text as Text
+import UIExplorer exposing (storiesOf)
+import Utils exposing (iconsSvgSprite, storyWithModel)
+
+
+update : Checkboxes.Msg -> Checkboxes.Model -> Return Checkboxes.Msg Checkboxes.Model
+update msg model =
+    case msg of
+        Checkboxes.Box1Set newValue ->
+            ( { model | box1 = newValue }, Cmd.none )
+
+        Checkboxes.Box2Set newValue ->
+            ( { model | box2 = newValue }, Cmd.none )
+
+        Checkboxes.Box3Check ->
+            ( { model | box3 = True }, Cmd.none )
+
+        Checkboxes.Box3Uncheck ->
+            ( { model | box3 = False }, Cmd.none )
+
+
+stories renderConfig =
+    storiesOf
+        "Checkboxes"
+        [ boxDemo renderConfig ]
+
+
+boxDemo renderConfig =
+    let
+        implicitMessages newState =
+            if newState then
+                Checkboxes.Box3Check
+
+            else
+                Checkboxes.Box3Uncheck
+
+        view { checkboxesStories } =
+            Element.column [ Element.spacing 8 ]
+                [ iconsSvgSprite
+                , Checkbox.checkbox renderConfig
+                    (Checkboxes.Box1Set >> Msg.CheckboxesStoriesMsg)
+                    "Extra ketchup (+0.50 USD)"
+                    checkboxesStories.box1
+                , Checkbox.checkbox renderConfig
+                    (Checkboxes.Box2Set >> Msg.CheckboxesStoriesMsg)
+                    "Large french fries (+0.50 USD)"
+                    checkboxesStories.box2
+                , Checkbox.checkbox renderConfig
+                    (implicitMessages >> Msg.CheckboxesStoriesMsg)
+                    "Heinz® Mayonnaise (+0.75 USD)"
+                    checkboxesStories.box3
+                , ("Total: "
+                    ++ (String.fromFloat <| totalPrice checkboxesStories)
+                    ++ " USD"
+                  )
+                    |> Text.body1
+                    |> Text.renderElement renderConfig
+                ]
+
+        totalPrice { box2, box3 } =
+            List.sum
+                [ 4.8
+                , ifThenElse box2 0.5 0
+                , ifThenElse box3 0.75 0
+                ]
+    in
+    storyWithModel ( "Checkboxes", view, { note = """
+This demo can be reproduced with the following code:
+
+```elm
+    Element.column [ Element.spacing 8 ]
+        [ Checkbox.checkbox renderConfig
+            Msg.Box1Set
+            "Extra ketchup (+0.50 USD)"
+            model.box1
+        , Checkbox.checkbox renderConfig
+            Msg.Box2Set
+            "Large french fries (+0.50 USD)"
+            model.box2
+        , Checkbox.checkbox renderConfig
+            (\\newState -> ifThenElse newState Msg.Box3Check Msg.Box3Uncheck)
+            "Heinz® Mayonnaise (+0.75 USD)"
+            model.box3
+        , ("Total: "
+            ++ (String.fromFloat <| totalPrice model)
+            ++ " USD"
+          )
+            |> Text.body1
+            |> Text.renderElement renderConfig
+        ]
+```""" } )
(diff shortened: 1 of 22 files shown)


src/UI/Checkbox.elm
```diff
@@ -0,0 +1,79 @@
...
+        boxAttrs =
+            [ Element.width (px 26)
+            , Element.height (px 26)
+            , Border.color Palette.primary.middle
+            , Border.width 3
+            , Border.rounded 8
+            , Events.onClick (message (not state))
+            , Element.pointer
+            ]
+                ++ (ARIA.toElementAttributes <| ARIA.roleButton)
```
jouderianjr, line 40: the checkbox is not a button, check the accessibility attributes: [click here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role)

src/UI/Internal/FiltersHeaders.elm
```diff
@@ -87,7 +106,7 @@ headerApplied : RenderConfig -> msg -> msg -> String -> String -> Element msg
 headerApplied renderConfig openMsg clearMsg clearHint label =
     -- Button.primary
     Element.row (Element.onIndividualClick openMsg :: headerAttrs True)
-        [ Element.text label
+        [ Element.text (label ++ " (1)")
```
jouderianjr, line 109: I didn't get this, what is this?

src/UI/Internal/FiltersHeaders.elm
```diff
@@ -125,6 +144,8 @@ headerAttrs isApplied =
             else
                 [ Background.color Palette.gray.lightest
                 , Font.color Palette.primary.darkest
+                , Element.mouseOver
+                    [ Background.color Palette.primary.lightest ]
```
jouderianjr, line 148: I would use the same smooth transition we have for the buttons, but it's up to you

src/UI/Internal/FiltersHeaders.elm
```diff
@@ -287,6 +308,180 @@ singleTextFilterRender renderConfig applyMsg { fromFiltersMsg, index, label } ed
...
+    -> msg
+    -> Config msg
+    -> Filters.Editable (Array String)
+    -> Element msg
+multiTextFilterRender renderConfig applyMsg { fromFiltersMsg, index, label } editableArr =
+    let
+        editMsg subIndex str =
+            fromFiltersMsg <| Filters.EditMultiText { column = index, field = subIndex, value = str }
+
+        single subIndex line =
```
jouderianjr, line 328: I would try to use more explcit names, avoiding too generic names, `single` can be a lot of things, so maybe `textField` or `singleTextField` something like this.

src/UI/Internal/FiltersHeaders.elm
```diff
@@ -287,6 +308,180 @@ singleTextFilterRender renderConfig applyMsg { fromFiltersMsg, index, label } ed
...
+                |> TextField.singlelineText (editMsg subIndex) label
+                |> TextField.withSize Size.extraSmall
+                |> TextField.withWidth TextField.widthFull
+                |> TextField.withOnEnterPressed applyMsg
+                |> TextField.renderElement renderConfig
+
+        instanceArr =
+            Filters.editableWithDefault Array.empty editableArr
+    in
+    instanceArr
```
jouderianjr, line 339: I would remove instanceArr function and just:

```elm
editableArry
        |> Filters.editableWithDefault Array.empty
        |> Array.push ""
        |> Array.indexedMap single
        |> Array.toList
        |> Element.column [ Element.width fill, Element.spacing 8 ]
```
just to reduce the let..in functions

src/UI/Internal/FiltersHeaders.elm
```diff
@@ -287,6 +308,180 @@ singleTextFilterRender renderConfig applyMsg { fromFiltersMsg, index, label } ed
...
+    -> Element msg
+selectFilterRender renderConfig { fromFiltersMsg, index } list { current, applied } =
+    let
+        editMsg subIndex =
+            fromFiltersMsg <| Filters.EditSelect { column = index, value = subIndex }
+
+        selected =
+            maybeNotThen applied current
+    in
+    radioGroup renderConfig
```
jouderianjr, line 360: I would use `radioGroup` qualified, to be more explicit that's an external funcion.

src/UI/RadioButton.elm
```diff
@@ -0,0 +1,96 @@
+module UI.RadioButton exposing (radioGroup)
```
jouderianjr, line 1: Apparently, this is called only `radio` everywhere, ( I checked all the places). so, I would rename it to `UI.radio`

src/UI/RadioButton.elm
```diff
@@ -0,0 +1,96 @@
...
+import UI.Icon as Icon
+import UI.Internal.Palette as Palette
+import UI.Palette as Palette
+import UI.RenderConfig exposing (RenderConfig)
+import UI.Text as Text
+import UI.Utils.ARIA as ARIA
+import UI.Utils.Element as Element
+
+
+radioGroup : RenderConfig -> String -> (a -> msg) -> Maybe a -> List ( a, String ) -> Element msg
```
jouderianjr, line 25: we can call this only group right? then often will be used like `Radio.group`

src/UI/RadioButton.elm
```diff
@@ -0,0 +1,96 @@
...
+        radioAttrs =
+            [ Element.width (px 26)
+            , Element.height (px 26)
+            , Border.color Palette.primary.middle
+            , Border.width 3
+            , Border.rounded 26
+            , Events.onClick message
+            , Element.pointer
+            ]
+                ++ (ARIA.toElementAttributes <| ARIA.rolePresentation)
```
jouderianjr, line 53: I think it's `role="radio"`  here

showcase/src/Main.elm
```diff
@@ -18,6 +18,7 @@ import Sizes
 import Tables.Stories as Tables
 import TextField
 import Texts
+import UI.Checkbox as Checkbox
```
jouderianjr, line 21: is this being used?

PedroHLC: No, it's just here so we know when compilations fail. There isn't yet a single use-case for it, just the component on Design Core System's Figma.

src/UI/Checkbox.elm
PedroHLC: You're right, fixing it...

src/UI/Internal/FiltersHeaders.elm
PedroHLC: Me neither, I just saw it yesterday on Figma and didn't ask Sufyian yet.
It can be seen in "Col Header - FilterApplied, CTA Hovered"  on Figma.

src/UI/Internal/FiltersHeaders.elm
PedroHLC: You're right, applying it...

src/UI/RadioButton.elm
PedroHLC: Nope, according to the MDN/W3C:
The most extern column is the radiogroup
The row that has both icons and text is the radio (and has an aria-checked)
The image/icon/symbol/figure that represents the current state must be ignored as we already have it stated in aria-checked, so it's a presentation/none.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_radio_role

PedroHLC:
There aren't use-cases for Checkboxes yet on Figma, not even as table filters. Just the component on Design Core System's Figma.
But yeah, I'll add them to the Showcase's stories.

src/UI/RadioButton.elm
jouderianjr: that's true, my bad

showcase/src/Checkboxes/Stories.elm
```diff
@@ -0,0 +1,100 @@
...
+                , ("Total: "
+                    ++ (String.fromFloat <| totalPrice checkboxesStories)
+                    ++ " USD"
+                  )
+                    |> Text.body1
+                    |> Text.renderElement renderConfig
+                ]
+
+        totalPrice { box2, box3 } =
+            List.sum
```
jouderianjr, line 70: you forgot about the poor box1, he/she has feelings man!

PedroHLC: Extra ketchup is for free!

PR #62, opened 2020-07-23
WMS-399 | Fix ellipsis behaviour

jouderianjr:
So apparently, the ellipsis feature was not working properly, I'm not sure if ever did taking into account this issue opened: [Click here](https://github.com/mdgriffith/elm-ui/issues/49).

Before: 

Screenshot 2020-07-23 at 19 53 42

After

Screenshot 2020-07-23 at 19 54 14

It's quite hackish, but blame elm-ui :o

### DIFF
--- a/src/UI/Internal/Text.elm
+++ b/src/UI/Internal/Text.elm
@@ -3,7 +3,6 @@ module UI.Internal.Text exposing (..)
 import Element exposing (Attribute, Element, fill)
 import Element.Font as Font
 import Html
-import Html.Attributes
 import List
 import UI.Internal.Basics exposing (ifThenElse)
 import UI.Internal.Palette as Palette
--- a/src/UI/Internal/Text.elm
+++ b/src/UI/Internal/Text.elm
@@ -6,6 +6,7 @@ import Html
 import List
 import UI.Internal.Basics exposing (ifThenElse)
 import UI.Internal.Palette as Palette
+import UI.Internal.Utils.Element exposing (tuplesToStyles)
 import UI.Palette as Palette exposing (brightnessDarkest, toneGray)
 import UI.RenderConfig exposing (RenderConfig, isMobile)
 import UI.Utils.Element as Element
@@ -104,7 +105,7 @@ attributes config size ellipsis color =
             else
                 deskAttributes size
 
-        ellipsisAttrs attrs =
+        heightAttr attrs =
             if ellipsis then
                 oneLineHeight mobile size :: attrs
 
@@ -120,7 +121,7 @@ attributes config size ellipsis color =
                     attrs
     in
     sizeAttrs
-        |> ellipsisAttrs
+        |> heightAttr
         |> colorAttr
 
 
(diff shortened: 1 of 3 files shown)


src/UI/Utils/Element.elm
```diff
@@ -179,6 +179,8 @@ ellipsis : List (Attribute msg)
 ellipsis =
```
jouderianjr, line 179: I'm not sure if it makes sense to have this function "public" and also outside of Internal.Text 🤔

jouderianjr:
@PedroHLC I've built testing in wms, but I will do another check in all the screens :D

src/UI/Internal/Utils/Element.elm
```diff
@@ -0,0 +1,14 @@
...
+import Element exposing (Attribute)
+import Html.Attributes as HtmlAttrs
+
+
+style : String -> String -> Attribute msg
+style k v =
+    Element.htmlAttribute <| HtmlAttrs.style k v
+
+
+tuplesToStyles : ( String, String ) -> Attribute msg
```
jouderianjr, line 12: I renamed this function to be more explicit about what it does. WDYT?

src/UI/Internal/Text.elm
```diff
@@ -365,3 +367,28 @@ combinedAttrs { ellipsis } =
...
+            (ellipsisAttrs lineHeightSize)
+
+
+ellipsisAttrs : Int -> List (Attribute msg)
+ellipsisAttrs lineHeightSize =
+    [ ( "text-overflow", "ellipsis" )
+    , ( "white-space", "nowrap" )
+    , ( "overflow", "hidden" )
+    , ( "display", "block" )
+    , ( "line-height", String.fromInt lineHeightSize ++ "px" )
```
jouderianjr, line 391: I had to also add this here, otherwise, the text will be cut for some TextSizes

src/UI/Internal/Text.elm
```diff
@@ -365,3 +367,28 @@ combinedAttrs { ellipsis } =
...
+            lineHeight (isMobile cfg) size
+    in
+    content
+        |> Html.text
+        |> Element.html
+        |> Element.el
+            (ellipsisAttrs lineHeightSize)
+
+
+ellipsisAttrs : Int -> List (Attribute msg)
```
jouderianjr, line 385: I don't think this should be external, as we should not use text without use `UI.Text`, if you think it makes sense be internally exposed, I can move to Internal.Utils.Element

PedroHLC: There are a lot of texts in internal components that don't fit any of our styles, like Filter's Header, Buttons, etc;

src/UI/Internal/Text.elm
PedroHLC: nice, I think it makes sense!

src/UI/Internal/Utils/Element.elm
PedroHLC: That's a pair, we may end up having a triple later.

PR #65, opened 2020-07-27
POC | Generate Elm code from i18n's JSONs

PedroHLC:
:tophat: What?
This PR adds I18n, generating Elm sources from JSONs using NodeJS.

:tophat: Why?
We can still manage the translations in POEditor, while having the safety of elm, and not requiring the app to redistribute and decode that JSON.

:clipboard: Subtasks
None


package.json
```diff
@@ -34,5 +35,7 @@
     "parcel-bundler": "^1.12.4",
     "parcel-plugin-asset-copier": "^1.1.0"
   },
-  "devDependencies": {}
+  "devDependencies": {
+    "parcel-plugin-elm-i18n": "PedroHLC/parcel-plugin-elm-i18n"
```
jouderianjr, line 39: Does it make sense to call it like this if we won't use it as a parcel plugin?

package.json
```diff
@@ -6,13 +6,14 @@
...
-  "assetsPath" : "showcase/public/assets/",
+  "assetsPath": "showcase/public/assets/",
   "scripts": {
     "analyse": "elm-analyse",
     "docs": "elm make --docs=dist/docs.json",
     "build": "parcel build showcase/public/index.html --public-url ./",
     "showcase": "parcel showcase/public/index.html",
-    "test": "elm-test"
+    "test": "elm-test",
+    "i18n": "node -e 'require(\"./node_modules/parcel-plugin-elm-i18n/index\")()' && elm-format --yes src/I18n"
```
jouderianjr, line 16: That's a bit hackish I would say, you can create runnable commands in the `parcel-plugin-elm-i18n` to make it runnable as something like: `npx convert-i18n-to-elm` or something like that.

Also, I think this elm-format should be handled by the library and not here 🤔

src/I18n/Type.elm
```diff
@@ -0,0 +1,49 @@
+module I18n.Type exposing (..)
```
jouderianjr, line 1: `Types` maybe?

src/UI/Internal/RenderConfig.elm
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+
+{-| `RenderConfig.RenderConfig` upholds all the information required for the components to apply the responsivity and accessibility changes.
+-}
+type RenderConfig
+    = RenderConfig RenderConfigData
+
+
+type Locale
```
jouderianjr, line 22: We do not have any distinction between south American Spanish and Spanish from Spain, same for English, so I would call it just as English | Spanish

src/UI/Internal/RenderConfig.elm
```diff
@@ -0,0 +1,39 @@
...
+type alias RenderConfigData =
+    { deviceClass : Element.DeviceClass
+    , deviceOrientation : Element.Orientation
+    , locale : Locale
+    }
+
+
+{-| `RenderConfig.RenderConfig` upholds all the information required for the components to apply the responsivity and accessibility changes.
+-}
+type RenderConfig
```
jouderianjr, line 18: I would not make RenderConfig opaque internally, only inside of `UI.RenderConfig.elm`, this way I think it's easy to work with internally.

src/UI/Internal/RenderConfig.elm
```diff
@@ -0,0 +1,39 @@
...
+    = RenderConfig RenderConfigData
+
+
+type Locale
+    = EnglishGB
+    | SpanishES
+
+
+string : (I18n.Root -> String) -> RenderConfig -> String
```
jouderianjr, line 27: Is this function used?

src/UI/Internal/RenderConfig.elm
```diff
@@ -0,0 +1,39 @@
...
+string getter cfg =
+    getter (strings cfg)
+
+
+strings : RenderConfig -> I18n.Root
```
jouderianjr, line 32: I think I prefer the word term/terms here 🤔  as string/strings are quite generic and may be confusing depending of the context we will use.

jouderianjr: or maybe localeStrings/localeTerms

package.json
jouderianjr: do you think it's beneficial having it as an external dependency instead of having as a script inside this repo?

src/UI/Internal/RenderConfig.elm
PedroHLC: :thinking: we have one detail, our English use dates in the format "DD/MM/YYYY".

src/UI/Internal/RenderConfig.elm
PedroHLC: It would be impossible for us to extract it...

src/UI/Internal/RenderConfig.elm
PedroHLC: No, this was a proposal of another syntax I intended to do for you to review, but it seems worst:
`(RenderConfig.string (.fields >> .range >> .to) renderConfig)`
instead of
`(RenderConfig.strings renderConfig).fields.range.to`

WDYT?

src/UI/Internal/RenderConfig.elm
PedroHLC: too much pollution when writing:
`(RenderConfig.localeStrings renderConfig).fields.period.before`

PedroHLC:
I've abused the fact that `>>` has more precedence than `|>` and made it this standard:
`(renderConfig |> localeTerms >> .filters >> .select >> .description)`

It's the only one that met my eyes reading requirements...

i18n/Spanish.json
```diff
@@ -0,0 +1 @@
+English.json
```
jouderianjr, line 1: I've added a project in POEditor and tagged Luis there, so I hope we can get the translations soon.

src/UI/Internal/DateInput.elm
```diff
@@ -162,16 +164,19 @@ ifValid if_ then_ value =
...
+toTextField : RenderConfig -> String -> (String -> msg) -> String -> DateInput -> TextField msg
+toTextField cfg separator editMsg label current =
     let
         correct =
             current
                 |> toDD_MM_YYYY separator
                 |> TextField.singlelineText editMsg label
 
+        invalidTerm =
+            cfg |> localeTerms >> .dateInput >> .invalid
```
jouderianjr, line 176: To be honest, not sure if this is better, still quite long, but I'm okay with that anyway hahaha :D

src/UI/Internal/RenderConfig.elm
jouderianjr: Why? I didn't get it 🤔

src/UI/Internal/RenderConfig.elm
jouderianjr: Still don't think it's relevant enough to have this distinction here, once we have it, we do it, otherwise, let's use English and Spanish.

src/UI/Internal/RenderConfig.elm
PedroHLC: The other components are calling `localeTerms`, they have the opaque `RenderConfig`, I would need to add a method somewhere to "de-opaque" it for using `localeTerms`.

PR #66, opened 2020-07-28
Add PR template

jouderianjr:
Adding a PR template, the template is:

#### :thinking: What?
Describe here

#### :man_shrugging: Why?
Describe here

#### :pushpin: Jira Issue
[JIRA-CODE](jira-url)

#### :no_good: Blocked by
[JIRA-CODE](jira-url) or [PR#CODE](pr-url)

#### :clipboard: Pending
- [ ] Feed the trolls.

### :fire: Extra

### DIFF
--- a/.github/workflows/PULL_REQUEST_TEMPLATE.md
+++ b/.github/workflows/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,12 @@
+#### :thinking: What?
+Describe here
+
+
+#### :man_shrugging: Why?
+Describe here
+
+#### :pushpin: Related Issues
+- [JIRA-CODE](jira-url) 
+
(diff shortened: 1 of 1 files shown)


PedroHLC:
I don't see much how we're gonna use "Related Issues" as a list. It's good to have a hyperlink to the main JIRA task, but then from there, all the related cards should be already visible.

I would also vote for having a:
* Based on #66
(For those packages that aren't rebased to master yet)
* Blocked by rails-backend/#00
* Alternative implementation of #66
* Pending discussions

jouderianjr:
@PedroHLC I've added a topic called `Extra` to support the things you mentioned that are not required and don't appear every time, WDYT?

PedroHLC:
Release the Kraken!

PR #71, opened 2020-08-07
WMS-438 | Allow custom element table cells

segmentationfaulter:
#### :thinking: What?
Previously we could only render a button or text in table cells, this change will allow rendering any custom element.

#### :man_shrugging: Why?
[WMS-416](https://paacklogistics.atlassian.net/browse/WMS-416) needs to render a badge inside a cell, and waiting on this change to get merged.

#### :pushpin: Jira Issue
[WMS-438](https://paacklogistics.atlassian.net/browse/WMS-438)

#### :no_good: Blocked by
None

### DIFF
--- a/src/UI/Tables/Common.elm
+++ b/src/UI/Tables/Common.elm
@@ -143,7 +143,9 @@ cellFromButton text =
 
 
 {-| Creates a cell with an `Element msg` inside.
-    cellFromCustom <| Element.row [] [Element.text "Hello", Element.text "World"]
+
+    cellFromCustom <| Element.row [] [ Element.text "Hello", Element.text "World" ]
+
 -}
 cellFromCustom : Element msg -> Cell msg
 cellFromCustom element =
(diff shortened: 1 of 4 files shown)


src/UI/Tables/Common.elm
```diff
@@ -3,6 +3,7 @@ module UI.Tables.Common exposing
     , ColumnWidth, columnWidthPortion, columnWidthPixels
     , Row, ToRow, rowEmpty, rowCellText, rowCellButton
     , Cell, cellFromText, cellFromButton
+    , cellFromCustom, rowCellCustom
```
jouderianjr, line 6:
```suggestion
    , Row, ToRow, rowEmpty, rowCellText, rowCellButton, rowCellCustom
    , Cell, cellFromText, cellFromButton, cellFromCustom
```

This way we keep them grouped by context

PedroHLC: For the elm-format to leave it in this way, you must sort it to in the `@docs` and not here

segmentationfaulter: >For the elm-format to leave it in this way, you must sort it to in the @docs and not here

Thanks @PedroHLC for the tip, may I read some documentation how `@docs` work?

segmentationfaulter:
I have modified a column to showcase [usage of `rowCellCustom`](https://github.com/PaackEng/paack-ui/pull/71/commits/732519c7b95133b11ea0a5d6711a484fc778d54c), it looks ugly but serves the purpose I think. Let me know if you think we should do it differently?

cc: @jouderianjr @PedroHLC

PR #72, opened 2020-08-10
POC | Visual regression

renanpvaz:
#### :thinking: What?
The PR adds visual regression testing using Cypress + Percy

#### :man_shrugging: Why?
To ensure that changes causing visual regressions are not merged in

#### :clipboard: Pending
- [ ] Add PERCY_TOKEN env var.

### :fire: Extra
[Linking Percy to Github](https://docs.percy.io/docs/github)


cypress.json
```diff
@@ -0,0 +1,3 @@
+{
+  "baseUrl": "http://localhost:1234"
```
PedroHLC, line 2: LOL, can we serve from inside GH Actions? That's news for me...

renanpvaz: Actually this is for running it locally, we should be able to serve it but now that you brought this up I realized that the URL might be wrong, I'll try to run it locally to be sure

miguelcoba:
> This PR as it's seems okay to me.
> For the token, you're gonna need one of the repo's owners to add it. As Jouderian's on holidays, you'll probably have to ask murbano83.
> Have you got a "Paack" account for cypress? I think that's managed by the product team, maybe @miguelcoba can confirm that...
> 
> I find It cool even if we end up changing it for another service later...

I don't know if we have a paack account for cypress, @murbano83 or @h0lyalg0rithm should know or create one if needed.

renanpvaz:
@miguelcoba everything is working now, Suraj added the integration to the repo. Now if you look at the PR checks there's one for Percy, if you click "details" you can see the last build and all the snapshots it took

PR #73, opened 2020-08-10
WMS-415 | Ellipsize table cell content

PedroHLC:
#### :thinking: What?
* Wraps all `Table.cellText` with `Text.setEllipsis True`, adjusting padding for the new line-height.
* Add title to all ellipsed texts.

#### :pushpin: Jira Issue
[WMS-415](https://paacklogistics.atlassian.net/browse/WMS-415)

#### :no_good: Blocked by
Not blocked, based on master.

#### :clipboard: Pending
Nothing pending.

### :fire: Extra
I've experimented in making it optional and applying title in every text. But I think this is the best result.

### DIFF
--- a/src/UI/Internal/Utils/Element.elm
+++ b/src/UI/Internal/Utils/Element.elm
@@ -88,6 +88,16 @@ positionAbsolute =
     Element.htmlAttribute <| HtmlAttrs.style "position" "absolute"
 
 
+positionAbsoluteTop : Attribute ms
+positionAbsoluteTop =
+    Element.htmlAttribute <| HtmlAttrs.style "top" "1"
+
+
+positionAbsoluteLeft : Attribute ms
+positionAbsoluteLeft =
+    Element.htmlAttribute <| HtmlAttrs.style "left" "1"
+
+
 overflowVisible : Attribute msg
 overflowVisible =
     Element.htmlAttribute <| HtmlAttrs.style "overflow" "visible"
--- a/src/UI/Internal/Utils/Element.elm
+++ b/src/UI/Internal/Utils/Element.elm
@@ -61,6 +61,14 @@ overlay closeMsg content =
         (overlayBackground closeMsg)
 
 
+tabIndex : Int -> Attribute msg
+tabIndex code =
+    -- -1 is disabled, 0 is following document flow, greater than 0 are priorities numbers
+    String.fromInt code
+        |> HtmlAttrs.attribute "tabindex"
+        |> Element.htmlAttribute
+
+
 
 -- Some good-old CSS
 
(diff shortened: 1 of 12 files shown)


src/UI/Internal/Utils/Element.elm
```diff
@@ -24,3 +24,11 @@ ellipsisAttrs lineHeightSize =
...
         |> List.map tuplesToStyles
         |> (::) Element.clip
+        |> (::) (title titleContent)
+
+
+title : String -> Attribute msg
+title value =
+    value
+        |> HtmlAttrs.title
+        |> Element.htmlAttribute
```
renanpvaz, line 34: Asking as a doubt, do you guys have preference of avoiding point-free style functions for some reason?

PedroHLC: Jouderian usually finds them less readable, not on every case, but when it finds he comments on reviews.
I end up adopting on not using them, except on anonymous lambdas.

renanpvaz:
Just a thought, the title attribute is usually not recommended when it comes to accessibility. I suggest either a tooltip that also works when you click the cell or some other kind of interaction

PedroHLC:
This PR is going back to draft as Suffian suggested a better UX ([here](https://www.figma.com/proto/axQM7tfJ42Pr3rvG1IfVsD/Table-Ellipsis-UX?node-id=1%3A294&viewport=264%2C364%2C0.16660422086715698&scaling=min-zoom)) regarding @renanpvaz suggestion.

src/UI/Internal/Utils/Element.elm
miguelcoba: I strongly prefer to be explicit with arguments. It is easier to read by devs, but old and new, and don't require us to keep in mind which parameters are supposed to be there.
Also, is the responsibility of the compiler to detect this cases and optimize them accordingly. Write first for devs, then for compilers.

PedroHLC:
@renanpvaz Take a look at its current state, let me know if you have some better ideia. Because it became quite a complex change, but it's working on touch devices!

PedroHLC:
Preview:
Preview

PedroHLC:
**NOTE**: By now this change is demonstrated only in the author's column of the showcase table.

src/UI/Tables/Stateful.elm
```diff
@@ -417,6 +428,13 @@ updateSelectionSet state item value =
     )
 
 
+updateExpandEllipsis : StateModel msg item columns -> Int -> Int -> ( State msg item columns, Cmd msg )
+updateExpandEllipsis state row col =
+    ( State { state | expandedEllipsis = Just ( row, col ) }
+    , Cmd.none
+    )
+
```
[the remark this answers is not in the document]
PedroHLC: I indeed tried, but the focus in one element can't trigger the transparency change in one of its sub-components.
Besides that, a shenanigan is currently required so that a simple "el" is focusable.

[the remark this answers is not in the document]
renanpvaz: Maybe we could talk to Sufiyan about leaving the button always visible (TBH I think it would be better that way because we don't hide the action from the user). The issue with the transparency could be solved using a child selector I guess.
Anyway, it's up to you in the end, but I have the feeling that this might be too complex for a somewhat ordinary ui element. There are implementations of tooltips that only use css, something on the lines of:
```css
[data-tooltip] {
  content: attr('data-tooltip');
}

[data-tooltip]:after,
[data-tooltip]:before {
  // ...
}
```

PedroHLC:
Ok, now the only difference btw Figma and this one, is that there is no middle-state.
Hovering/focusing on the cell already shows the tooltip.
I've also added it to the Title column.

PR #76, opened 2020-08-18
Add min width to portion cells and examples

renanpvaz:
#### :thinking: What?
Fixing the issue where the portion columns were overflowing
Adding an example to cover this use case

#### :pushpin: Jira Issue
[WMS-473](https://paacklogistics.atlassian.net/browse/WMS-473)

### DIFF
--- a/cypress/integration/visual-regression.spec.js
+++ b/cypress/integration/visual-regression.spec.js
@@ -64,9 +64,26 @@ describe('Percy Snapshots', () => {
   })
 
   describe('Complex components', () => {
-    it('Creates the Tables page snapshot', () => {
-      cy.visit('/#Complex%20components/Tables/Desktop')
-      cy.percySnapshot('Tables')
+    describe('Tables', () => {
+      it('Creates the Desktop Tables page snapshot', () => {
+        cy.visit('/#Complex%20components/Tables/Desktop')
+        cy.percySnapshot('Tables Desktop')
+      })
+
+      it('Creates the Mobile Tables page snapshot', () => {
+        cy.visit('/#Complex%20components/Tables/Mobile')
+        cy.percySnapshot('Tables Mobile')
+      })
+
+      it('Creates the Stateless Tables page snapshot', () => {
+        cy.visit('/#Complex%20components/Tables/Stateless')
+        cy.percySnapshot('Tables Stateless')
+      })
+
+      it('Creates the Portion Column Tables page snapshot', () => {
+        cy.visit('/#Complex%20components/Tables/Portion%20Column')
+        cy.percySnapshot('Tables Portion Column')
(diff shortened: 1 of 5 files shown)


cypress/integration/visual-regression.spec.js
```diff
@@ -64,9 +64,26 @@ describe('Percy Snapshots', () => {
...
+
+      it('Creates the Stateless Tables page snapshot', () => {
+        cy.visit('/#Complex%20components/Tables/Stateless')
+        cy.percySnapshot('Tables Stateless')
+      })
+
+      it('Creates the Portion Column Tables page snapshot', () => {
+        cy.visit('/#Complex%20components/Tables/Portion%20Column')
+        cy.percySnapshot('Tables Portion Column')
+      })
```
renanpvaz, line 86: should we continue to add every case like this or does it make sense to make something that visits all pages and stories creating snapshots?

jouderianjr: I liked it this way, seems good to me. WDYT?

showcase/src/Tables/Stories.elm
```diff
@@ -114,7 +115,17 @@ statelessDemoTable renderConfig =
 selectableTableStory renderConfig =
     storyWithModel
         ( "Selectable"
-        , \{ tablesStories } -> withIcons renderConfig (demoTable renderConfig tablesStories.selecTableState Stories.ForSelectable)
+        , \{ tablesStories } -> withIcons renderConfig (demoTable renderConfig Book.tablePixelColumns tablesStories.selecTableState Stories.ForSelectable)
```
jouderianjr, line 118: I would break this line, it is too long I think

showcase/src/Tables/Stories.elm
```diff
@@ -114,7 +115,17 @@ statelessDemoTable renderConfig =
...
+        , { defaultWithMenu
+            | note = "See [docs](https://package.elm-lang.org/packages/PaackEng/paack-ui/latest/UI-Tables-Stateful) for the exact code of this example."
+          }
+        )
+
+
+portionColumnsTable renderConfig =
+    storyWithModel
+        ( "Portion Columns"
+        , \{ tablesStories } -> withIcons renderConfig (demoTable renderConfig Book.tablePortionColumns tablesStories.selecTableState Stories.ForSelectable)
```
jouderianjr, line 128: Same here

jouderianjr (approved):

cypress/integration/visual-regression.spec.js
PedroHLC: :smile: I don't have an opinion right now, but I think this discussion should be moved to its own PR as this fix is quite urgent.

PR #81, opened 2020-08-26
WMS-506 improve table tooltip

renanpvaz:
Captura de Tela 2020-08-26 às 5 11 18 PM

#### :thinking: What?
Creating a custom element that detects when an overflow happens and displays a tooltip underneath it

#### :man_shrugging: Why?
Because the previous Elm implementation was not able to detect it

#### :pushpin: Jira Issue
[WMS-506](https://paacklogistics.atlassian.net/browse/WMS-506)


showcase/public/index.js
```diff
@@ -1,6 +1,100 @@
...
+
+      tooltip.setAttribute('tabIndex', 0)
+
+      tooltip.className = 'tooltip'
+      this.shadowRoot.appendChild(tooltip)
+    }
+  }
+}
+
+window.customElements.define('ellipsizable-text', EllipsizableText)
```
renanpvaz, line 96: This will be moved to `paack-ui-assets` once https://github.com/PaackEng/paack-ui-assets/pull/1 is merged

PedroHLC: In the while, can you keep it in another js file, so it makes it easier to review like if it was already moved?

showcase/src/Tables/Book.elm
```diff
@@ -45,7 +45,7 @@ someFilters =
 
 
 books =
-    [ { author = "Dan Brown"
+    [ { author = "Dan Lorem ipsum dolor sit amet consectetur adipisicing elit. "
```
PedroHLC, line 48: Don't break my heart like this, there is the Narnia book, just change the default filter to always show it! :sweat_smile:

renanpvaz: sorry this wasn't meant to be pushed lol

showcase/public/ellipsizable-text.js
```diff
@@ -0,0 +1,97 @@
+const template = Object.assign(document.createElement('template'), {
```
jouderianjr, line 1: So, I was doing some tests here because I noticed that we are adding the ellipsis class only when we know that the text should overflow, this generates a redraw. we can set the ellipsis attributes by default and the just calculate if we should show the tooltip or not, this way we avoid the redraw right?

showcase/public/ellipsizable-text.js
```diff
@@ -0,0 +1,97 @@
...
+        .text {
+          white-space: nowrap;
+        }
+  
+        .text--overflown {
+          cursor: pointer;
+          text-overflow: ellipsis;
+          overflow: hidden;
+          display: block;
+        }
```
jouderianjr, line 13:
```css
        .text {
          white-space: nowrap;
          text-overflow: ellipsis;
          overflow: hidden;
          display: block;
        }
  
        .text--overflown {
          cursor: pointer;
        }
```

showcase/public/ellipsizable-text.js
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  connectedCallback() {
+    const { textContent } = this
+    const text = this.shadowRoot.querySelector('.text')
+
+    text.textContent = textContent
+
+    runDelayed(() => {
+      if (text.offsetWidth > this.parentElement.offsetWidth) {
```
jouderianjr, line 82: I found out that you can do the if condition like this:

`if (text.offsetWidth < text.scrollWidth) {` 

this way you don't need to access parentElement.

showcase/public/index.js
```diff
@@ -1,6 +1,7 @@
-import { Elm } from './../src/Main.elm';
-import paackSvgIconSprite from 'paack-ui-assets/js/paackSvgIconSprite';
+import { Elm } from './../src/Main.elm'
```
jouderianjr, line 1: Probably this is by default, but you changed the code style here( removing the `;` everywhere) I think we should leave as it was (also change your file to contain `;`) and then in another pr/discussion we can decide in a JS code style and add the config file to the project, WDYT?

src/UI/Internal/Tables/View.elm
```diff
@@ -32,7 +31,15 @@ cellContentRender renderConfig cell_ =
 simpleText : RenderConfig -> Text -> Element msg
 simpleText renderConfig text =
     text
-        |> EllipsizableTooltip.view renderConfig
+        |> Text.setEllipsis True
```
jouderianjr, line 34: Should we use Keyed.el here in order to avoid unnecessary redraws? 🤔

jouderianjr (commented):
I found a bug when the text is too long:

Screen Shot 2020-08-27 at 13 48 41

showcase/public/ellipsizable-text.js
renanpvaz: we can move some of these properties but not all, if I do that then when I get the width it won't be more than the parent's

showcase/public/ellipsizable-text.js
renanpvaz: I'm not sure we can avoid the redraw, it happens because I first render just the plain text letting it overflow and **then** I truncate it. If we skip the overflow then we can't detect it because the text's width will always match its parent. The only alternatives to detect the overflow without using `offsetWidth` are either using a canvas with `measureText` (which I'm pretty sure is slower) and using a monospaced font for the table so we can count the cols.

src/UI/Internal/Tables/View.elm
PedroHLC: wouldn't work, the cols/rows until here should have keyed while this one should use Lazy.el

jouderianjr (approved):

PR #86, opened 2020-09-01
Replace setEllipsis with withOverflow

renanpvaz:
#### :thinking: What?
Replace `setEllipsis` function with a new `withOverflow` function

#### :man_shrugging: Why?
Because we need to stop defaulting to the tooltip behavior for every case

#### :pushpin: Jira Issue
[WMS-519](https://paacklogistics.atlassian.net/browse/WMS-519)

### DIFF
--- a/src/UI/Internal/Text.elm
+++ b/src/UI/Internal/Text.elm
@@ -6,7 +6,7 @@ import Html exposing (Html)
 import List
 import UI.Internal.Basics exposing (ifThenElse)
 import UI.Internal.Palette as Palette
-import UI.Internal.Utils.Element exposing (ellipsisAttrs, overflowVisible)
+import UI.Internal.Utils.Element exposing (oveflowAttrs, overflowVisible)
 import UI.Palette as Palette exposing (brightnessDarkest, toneGray)
 import UI.RenderConfig exposing (RenderConfig, isMobile)
 import UI.Utils.Element as Element
@@ -22,7 +22,6 @@ type Span
 
 type alias Options =
     { color : TextColor
-    , oneLineEllipsis : Bool
     }
 
 
@@ -33,7 +32,7 @@ type alias SpanProperties =
 
 
 type alias TextOptions =
-    { ellipsis : Bool }
+    { overflow : TextOverflow }
 
 
 type TextSize
@@ -57,6 +56,12 @@ type TextColor
     | ColorInherit
 
 
+type TextOverflow
+    = Ellipsize
+    | EllipsizeWithTooltip
+    | Wrap
+
+
 defaultText : TextSize -> String -> Text
 defaultText size content =
     Text [ Span (SpanProperties content size) spanDefaultOptions ] textDefaultOptions
@@ -65,13 +70,12 @@ defaultText size content =
 spanDefaultOptions : Options
 spanDefaultOptions =
     { color = defaultColor
-    , oneLineEllipsis = False
     }
 
 
 textDefaultOptions : TextOptions
 textDefaultOptions =
-    { ellipsis = False }
+    { overflow = Wrap }
 
 
 defaultColor : TextColor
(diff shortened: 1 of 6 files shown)


src/UI/Internal/Text.elm
```diff
@@ -360,7 +360,7 @@ renderText cfg overflow size text =
             ellipsizedTextWith cfg
                 ellipsizableNode
                 [ ( "white-space", "normal" )
-                , ( "oveflow", "visible" )
```
PedroHLC: :scream:  how did this passed review? :joy:

src/UI/Text.elm
```diff
@@ -60,9 +61,9 @@ A text can be created and rendered as in the following pipeline:
 @docs withColor
 
 
-# Ellipsis
+# Overflow
 
-@docs setEllipsis
+@docs withOverflow
```
jouderianjr, line 66: Can you add the other related functions here as well? (ellipsize, ellipsizeWithTooltip, and wrap)

src/UI/Internal/Text.elm
```diff
@@ -57,6 +56,12 @@ type TextColor
     | ColorInherit
 
 
+type TextOverflow
+    = Ellipsize
+    | EllipsizeWithTooltip
+    | Wrap
```
PedroHLC, line 62: I see there isn't a `NoWrap` or `BreakWords`, but that's okay by now, when the need appears we add them

src/UI/Text.elm
```diff
@@ -183,18 +184,40 @@ withColor color text =
...
+
+{-| Lets the text break lines to prevent overflow.
+Default behavior.
+-}
+wrap : TextOverflow
+wrap =
+    Internal.Wrap
+
+
+{-| Determinines how the text overflow is handled.
```
jouderianjr, line 209: typo here

src/UI/Internal/Utils/Element.elm
```diff
@@ -16,13 +16,13 @@ tuplesToStyles ( k, v ) =
...
-ellipsisAttrs lineHeightSize titleContent =
-    [ ( "overflow", "visible" )
-    , ( "display", "block" )
-    , ( "white-space", "normal" )
-    , ( "line-height", String.fromInt lineHeightSize ++ "px" )
-    ]
+overflowAttrs : Int -> List ( String, String ) -> String -> List (Attribute msg)
+overflowAttrs lineHeightSize attrs titleContent =
+    attrs
+        |> (++)
```
PedroHLC, line 22: :thinking: `(++)` is an O(n) operation, It's better to do it in:
```
|> (::) ("display", "block") 
|> (::) ("line-height", String.fromInt lineHeightSize ++ "px") 
```
These are O(1) operations (and AOT reductive, not sure if Elm will do it tought)!

src/UI/Internal/Text.elm
```diff
@@ -334,13 +338,32 @@ spanMapOptions applier (Span prop opt) =
...
+renderText cfg overflow size text =
+    case overflow of
+        Wrap ->
+            Element.text text
+
+        Ellipsize ->
+            ellipsizedText cfg size text
+
+        EllipsizeWithTooltip ->
+            ellipsizedTextWith cfg
```
jouderianjr, line 360: This function is only being used for the "ellipsize with tooltip" situation right? do you think it needs to be "generic"  like this instead of just be a specific function?

jouderianjr (changes requested):

miguelcoba (approved):

src/UI/Internal/Text.elm
renanpvaz: it's also used inside `ellipsizedText`, which is just a shorthand because it's still exposed and used in another instance. The difference between the two is the `(String -> Html msg)` function and the style props

jouderianjr: oh, true, I haven't realized, but then isn't better do create also a function ellipsizableWithTooltipText and use it here? I mean just to keep the consistency between both scenarios 🤔

jouderianjr (approved):

PR #87, opened 2020-09-01
WMS-518 | Fix filter dialog overflow and set dimensions for checkboxes and radio buttons

segmentationfaulter:
#### :thinking: What?
Makes table filters dialog to fit its contents, also sets the dimensions for radio buttons and checkboxes as per [feedback by Sufyan](https://www.figma.com/file/rqAoj2WwQZYoisP2G6d6nJ?node-id=1:41#37303481). I am making checboxes dimension changes part of this PR because they were contributing to the dialog overflow.

#### :man_shrugging: Why?
Since contents of table filter were overflowing its boundary sometimes, have a look at [this PR](https://github.com/PaackEng/wms-web/pull/114) for context.

#### :pushpin: Jira Issue
[WMS-518](https://paacklogistics.atlassian.net/browse/WMS-518)

### :fire: Extra
Now a filter dialog will be sometimes of lesser width than the column width. I tried setting a minimum width (equal to the column width) on the dialog but couldn't make it work because there is no function exposed by Elm-UI to convert `Element.Length` to pixels. I think having lesser width than the column width shouldn't be an issue as having more than column width is not an issue.

### DIFF
--- a/src/UI/Checkbox.elm
+++ b/src/UI/Checkbox.elm
@@ -183,13 +183,13 @@ pxSize : Size -> Int
 pxSize size =
     case size of
         Size.ExtraSmall ->
-            16
+            10
 
         Size.Small ->
-            20
+            16
 
         Size.Medium ->
-            26
+            20
 
         Size.Large ->
-            32
+            26
--- a/src/UI/Checkbox.elm
+++ b/src/UI/Checkbox.elm
@@ -1,7 +1,6 @@
 module UI.Checkbox exposing
     ( Checkbox, checkbox
     , withLabelVisible
-    , withSize
     , renderElement
     )
 
@@ -43,7 +42,6 @@ import Element.Events as Events
 import UI.Icon as Icon
 import UI.Internal.Palette as Palette
 import UI.Internal.RenderConfig exposing (localeTerms)
-import UI.Internal.Size as Size exposing (Size)
 import UI.Palette as Palette
 import UI.RenderConfig exposing (RenderConfig)
 import UI.Text as Text
@@ -66,7 +64,6 @@ type alias Properties msg =
 
 type alias Options =
     { labelVisible : Bool
-    , size : Size
     }
 
 
@@ -80,7 +77,7 @@ type alias Options =
 checkbox : String -> (Bool -> msg) -> Bool -> Checkbox msg
 checkbox label message state =
     Checkbox { message = message, label = label, state = state }
-        { labelVisible = True, size = Size.default }
+        { labelVisible = True }
 
 
 {-| Show or hide the checkbox's label.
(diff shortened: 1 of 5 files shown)


miguelcoba (approved):

src/UI/Internal/Tables/FiltersView.elm
```diff
@@ -297,7 +297,7 @@ dialog renderConfig config filter clearMsg applyMsg content =
     in
     overlay config.discardMsg <|
         Element.column
-            [ Element.width fill
+            [ Element.width <| Element.shrink
```
PedroHLC, line 300: How does this change affect the filters with just TextFields? Cause they're all using "width fill".
We may need a hackish solution where `min-width: 100%; width: min-content;`.
(I'm commenting this without testing, feel free to revoke If I'm wrong)

segmentationfaulter: >How does this change affect the filters with just TextFields? Cause they're all using "width fill".

This is how they look like:

Screenshot_2020-09-01 Storybook Elm

> We may need a hackish solution where min-width: 100%; width: min-content;.

is it possible to use raw css with elm-ui?

PedroHLC: Take a look at `UI.Internal.Utils.Element`

jouderianjr: > We may need a hackish solution where min-width: 100%; width: min-content;.

I think this can do the trick 🤔 

Also @segmentationfaulter I found something a bit odd and I'm not sure why exactly, but I think the size here shouldn't change right? 🤔 

Sep-01-2020 19-02-49

segmentationfaulter: > Also @segmentationfaulter I found something a bit odd and I'm not sure why exactly, but I think the size here shouldn't change right?

This is happening because as soon as we make a selection, red button to clear the filter appears at the bottom and dialog makes room for it due to the rule `width: min-content`. If dialog doesn't expand in width, red button will be slightly overflowing as seen in the gif below

dialog_overflow

> We may need a hackish solution where min-width: 100%; width: min-content;

Thanks @PedroHLC for the help, let me know if we need to change where we should place `dialogWidthAttrs`

jouderianjr: > This is happening because as soon as we make a selection, red button to clear the filter appears at the bottom and dialog makes room for it due to the rule width: min-content. If the dialog doesn't expand in width, red button will be slightly overflowing as seen in the gif below

got it, it makes sense :D. I'm wondering if does not make sense then to have a minimum width taking into account this both button size 🤔 , this way we avoid this resize.

Also, I think your changes made the radio really small, is that intended?

Screen Shot 2020-09-02 at 12 50 14

segmentationfaulter: >I'm wondering if does not make sense then to have a minimum width taking into account this both button size

I am afraid doing so could result in both the buttons overlapping in a situation when parent column width is too narrow.

>Also, I think your changes made the radio really small, is that intended?

This was basically a side effect of the changes I made with the sizes of checkboxes and radios as per [feedback by Sufyan](https://www.figma.com/file/rqAoj2WwQZYoisP2G6d6nJ?node-id=1:41#37303481). As per this change, all the checkboxes and radio buttons will be having a size of 20x20. I have pushed a fix for selectable table checboxes size though.

PedroHLC: So, wouldn't it make sense to remove `Checkbox.withSize` ?

segmentationfaulter: >So, wouldn't it make sense to remove Checkbox.withSize ?

Yes, it makes sense i think, should I go with it?

segmentationfaulter: @PedroHLC I have removed `Checkbox.withSize` in the latest commit, if we think they should be retained, i will revert the commit.

cc: @jouderianjr

src/UI/Checkbox.elm
```diff
@@ -66,7 +64,6 @@ type alias Properties msg =
 
 type alias Options =
```
segmentationfaulter, line 65: I was wondering if we should get rid of this record since it has only one field now and change the 'Checkbox msg` type like this:

```elm
type Checkbox msg = Checkbox (Properties msg) Bool
```

cc: @PedroHLC

PedroHLC: I would keep it just to maintain API pattern, but the rest of the team may not agree with me...

src/UI/Internal/Tables/FiltersView.elm
```diff
@@ -297,12 +297,13 @@ dialog renderConfig config filter clearMsg applyMsg content =
     in
     overlay config.discardMsg <|
         Element.column
-            [ Element.width fill
-            , zIndex 9
-            , Element.alignTop
-            , Palette.mainBackground
-            , Primitives.defaultRoundedBorders
-            ]
+            (List.append dialogWidthAttrs
```
renanpvaz, line 300: is this only used here? if so it might make sense to define it here as well. Also, I think `(++)` is more idiomatic, although in this case it probably makes more sense to use `(::)` with the two separate elements

segmentationfaulter: made this change

PR #90, opened 2020-09-02
 Use Keyed elements in Table

PedroHLC:
This PR is open for discussions as well!

#### :thinking: What?
Uses `Element.Keyed` in `UI.Table`, enhancing diffing.

#### :man_shrugging: Why?
I've made it as a POC, it ended up being easier then it looked like.

#### :gun:  Counter argument
As we don't have that many rows, it may not bring any benefit.

#### :pushpin: Jira Issue
None, mostly a POC.

#### :no_good: Blocked by
#89

#### :clipboard: Pending
In other PR:
- Do the same to `UI.ListView`
- Do the same to `UI.NavigationContainer`

### :fire: Extra
I watched more animes in 2020 than the rest of my entire life.

### DIFF
--- a/src/UI/Tables/Stateless.elm
+++ b/src/UI/Tables/Stateless.elm
@@ -57,6 +57,7 @@ Where `Book` is:
 -}
 
 import Element exposing (Element, shrink)
+import Element.Keyed as Keyed
 import UI.Internal.NArray as NArray
 import UI.Internal.Tables.Common exposing (..)
 import UI.Internal.Tables.View exposing (..)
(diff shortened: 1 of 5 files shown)


miguelcoba (approved):

renanpvaz (approved):

src/UI/Tables/Stateless.elm
```diff
@@ -187,11 +188,13 @@ desktopView renderConfig prop opt =
 headersRender :
     RenderConfig
     -> List Column
-    -> Element msg
+    -> ( String, Element msg )
 headersRender renderConfig columns =
-    Element.row
+    ( "@headers"
```
jouderianjr, line 193: that's a bit "dangerous" right? I mean, if we have a dynamic column title, this will prevent being redrawn right?

PedroHLC: Nope, this should cause the header row to be redrawn without redrawing the entire list (brothers and sisters), and only that...

jouderianjr (approved):

PR #95, opened 2020-09-09
WMS-549 create login layout module

renanpvaz:
#### :thinking: What?
Extract the login view from WMS to create a reusable "Auth" layout

#### :man_shrugging: Why?
Because we need a way to have the same look and feel for the BR editor's login screen as the WMS 

#### :pushpin: Jira Issue
[WMS-549](https://paacklogistics.atlassian.net/browse/WMS-549)

### DIFF
--- a/showcase/src/Layouts/Stories.elm
+++ b/showcase/src/Layouts/Stories.elm
@@ -1,47 +1,68 @@
 module Layouts.Stories exposing (stories, update)
 
+import Element exposing (Element)
 import Layouts.Model as LayoutsModel
 import Layouts.Msg as LayoutsMsg
+import Model exposing (Model)
 import Msg exposing (Msg)
-import PluginOptions exposing (defaultWithMenu)
+import PluginOptions exposing (defaultWithoutMenu)
 import Return exposing (Return)
 import UI.Badge as Badge exposing (Badge)
+import UI.Button as Button
 import UI.Layout.Auth as Auth
 import UI.RenderConfig exposing (RenderConfig)
 import UI.TextField as TextField
-import UI.Button as Button
 import UIExplorer exposing (storiesOf)
-import Utils exposing (ExplorerStory, ExplorerUI, goToDocsCallToAction, prettifyElmCode, storyList)
-import Element
+import Utils exposing (ExplorerStory, ExplorerUI, goToDocsCallToAction, prettifyElmCode, storyList, storyWithModel)
 
 
 update : LayoutsMsg.Msg -> LayoutsModel.Model -> Return LayoutsMsg.Msg LayoutsModel.Model
 update msg model =
-    ( {}, Cmd.none )
+    case msg of
+        LayoutsMsg.SetEmail e ->
+            ( { model | email = e }, Cmd.none )
+
+        LayoutsMsg.SetPassword p ->
+            ( { model | password = p }, Cmd.none )
+
+        LayoutsMsg.NoOp ->
+            ( model, Cmd.none )
 
 
 stories : RenderConfig -> ExplorerUI
 stories cfg =
-    let
-        msg =
-            Msg.LayoutsStoriesMsg LayoutsMsg.NoOp
-    in
     storiesOf
         "Layouts"
-        [ storyList
-            ( "Auth"
-            , [ Auth.view cfg
-                    { title = "Auth Layout Example Story"
-                    , logo = Element.image [] { src = "logo.png", description = "logo" }
-                    , emailField = TextField.username (always msg) "username" ""
-                    , passwordField = TextField.username (always msg) "username" ""
-                    , submitMsg = msg
-                    , submitButton = Button.fromLabel "Login" |> Button.cmd msg Button.primary
-                    }
-                ]
-            , { defaultWithMenu | code = code }
-            )
+        [ demo cfg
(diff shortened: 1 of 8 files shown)


showcase/src/Layouts/Stories.elm
```diff
@@ -0,0 +1,79 @@
...
+            Msg.LayoutsStoriesMsg
+    in
+    Auth.view renderConfig
+        { title = "Auth Layout Example Story"
+        , logo = Element.image [] { src = "logo.png", description = "logo" }
+        , emailField = TextField.username (msg << LayoutsMsg.SetEmail) "username" layoutsStories.email
+        , passwordField = TextField.username (msg << LayoutsMsg.SetPassword) "password" layoutsStories.password
+        , submitMsg = msg LayoutsMsg.NoOp
+        , submitButton = Button.fromLabel "Login" |> Button.cmd (msg LayoutsMsg.NoOp) Button.primary
+        }
```
renanpvaz, line 65: if you think that this is too constrained as a login page with too many config options and you rather have something more generic, another approach I imagined would be something more composable and less configurable like:

```elm
-- replace CenteredColumn with whatever you think sounds best
CenteredColumn.layout 
  [ CenteredColumn.header 
       [ -- render logo element
       , -- render title element
       , -- render anything really
       ]
  , CenteredColumn.content 
       [ TextField.username (msg << LayoutsMsg.SetEmail) "username" 
           |> TextField.withPlaceholder "jon@paack.co"
           |> TextField.setLabelVisible True
           |> TextField.withWidth TextField.widthFull
           |> TextField.renderElement renderConfig
        , TextField.username (msg << LayoutsMsg.SetPassword) "password"
           |> TextField.withPlaceholder "********"
           |> TextField.setLabelVisible True
           |> TextField.withWidth TextField.widthFull
           |> TextField.withOnEnterPressed submitMsg
           |> TextField.renderElement renderConfig
       ]
  , CenteredColumn.footer 
      [ Button.fromLabel "Login" 
           |> Button.cmd (msg LayoutsMsg.NoOp) Button.primary
           |> Button.renderElement renderConfig
      ]
  ]

File: showcase/src/Layouts/Stories.elm [hunk 2]
```diff
@@ -0,0 +1,78 @@
...
+import UI.RenderConfig exposing (RenderConfig)
+import UI.TextField as TextField
+import UIExplorer exposing (storiesOf)
+import Utils exposing (ExplorerStory, ExplorerUI, goToDocsCallToAction, prettifyElmCode, storyWithModel)
+
+
+update : LayoutsMsg.Msg -> LayoutsModel.Model -> Return LayoutsMsg.Msg LayoutsModel.Model
+update msg model =
+    case msg of
+        LayoutsMsg.SetEmail e ->
```
PedroHLC, line 21: I'm considering (and hesitating) moving this logic inside the component...

PedroHLC (approved):

showcase/src/Layouts/Stories.elm
jouderianjr: What do you think about doing something like:

```elm
    Auth.view renderConfig
        { title = "Auth Layout Example Story"
        , logo = Element.image [] { src = "logo.png", description = "logo" }
        , emailField = 
            { onTypeChanged = LayoutsMsg.SetEmail
            , placeholder = username
            , value = layoutsStories.email
            }
        , passwordField =
            { onTypeChanged = LayoutsMsg.SetPassword
            , placeholder = "password"
            , value = layoutsStories.password
            }
        , submitMsg = LayoutsMsg.NoOp
        , submitButton = { title = "Login", msg = LayoutsMsg.NoOp }
        , toMsg = Msg.LayoutsStoriesMsg
        }
```

this way we don't need to "map" all the places using `msg` and the UI is more strict, I think if the idea is to reuse the whole thing, let's think about what do we want to make it dynamic? As far as I can see we would only have the texts and images dynamic in this case.

also. just to close about the naming here, like I said before, I don't feel that the `SplitSelectable` and the Auth are in the same level of "componentization". That being said, what if we leave this one as UI.Pages and the SplitSelectable as UI.Layouts? I want to know everyone's opinion here as maybe I'm deadly wrong here :D cc @renanpvaz @PedroHLC @segmentationfaulter @miguelcoba.

renanpvaz: First on the msg thing, I'm not sure if that would be the case in other instances, I only had to map the messages myself because our stories don't do that, but in a real use case it would be a flat `SetEmail` msg for example and `toMsg` would have to be `always`.

The only thing I really want to reuse here is the disposition of the elements in WMS' login screen, I don't care what it's called. Personally, I don't see what's so different between this layout and the split selectable, one would be used in two places and the other in three. About complexity, both layouts require more or less the same amount of configuration. So again I'll argue that we just need better naming to avoid the confusion of this being a whole page with functionality.

I agree that having something constrained to render an email and a password field feels weird and too specific. With that in mind that I came up with the alternative above where instead of a config record we just take lists of `Element` as args. I'm not sure about the idea of a `UI.Pages` because this module doesn't include the login page functionality, so seems misleading when compared to WMS' idea of "pages".

renanpvaz:
The login screen was deprioritized so I'm closing this for now

PR #96, opened 2020-09-10
BRE-48 SplitSelectable Layout

renanpvaz:
#### :thinking: What?
Adding the SplitSelectable Layout and the SummaryListItem view along with a story displaying the usage

#### :man_shrugging: Why?
Because these components are needed for the main feature in the BR Editor

#### :pushpin: Jira Issue
[BRE-48](https://paacklogistics.atlassian.net/browse/BRE-48)

### :fire: Extra
The story only works for desktop right now because I couldn't work out the `Nav.Container` with the story view.


src/UI/NavigationContainer.elm
```diff
@@ -515,6 +515,26 @@ toBrowserDocument cfg page (Navigator model) =
...
+The result of this function is a ready-to-insert Elm UI's Element.
+
+**NOTE**: This is a way to work around the NavigationContainer to display only the content of the container. It's meant to be used in the project's showcase and only works for desktop.
+
+    someView
+        |> Nav.contentSingle
+        |> Nav.toShowcaseElement
+
+-}
+toShowcaseElement : Content msg -> Element msg
```
PedroHLC, line 528: I don't like this, not even a bit.
What if... we move "Content" to an Internal type, with this... and then use this internal in the showcase?

PedroHLC (approved):

PedroHLC (approved):

showcase/src/Layouts/Stories.elm
```diff
@@ -0,0 +1,122 @@
...
+stories cfg =
+    storiesOf
+        "Layouts"
+        [ demo cfg
+        ]
+
+
+demo : RenderConfig -> ExplorerStory
+demo renderConfig =
+    storyWithModel
```
jouderianjr, line 42: Does it make sense to add a border color or something like that just to indicate what is the exact component? Maybe it could help people that are not familiar with this UI

src/UI/SummaryListItem.elm
```diff
@@ -0,0 +1,85 @@
+module UI.SummaryListItem exposing (view)
```
jouderianjr, line 1: Do you guys think it's a good name? IMO is not very explicit what this does, but maybe is just me, WYDT? @PedroHLC @renanpvaz

renanpvaz: not sure how I feel about it, but I can't think of an alternative really

showcase/src/Layouts/Stories.elm
renanpvaz: to `storyWithModel` or only this story?

src/UI/SummaryListItem.elm
jouderianjr: Leave as is now then :D if we come up with a better name we  change in the future

showcase/src/Layouts/Stories.elm
jouderianjr: Well, that's a quite good question, I this problem affects more layouts, "big area" components", so maybe we could create another storyWithModel function with the border or adding a flag to this function enabling/disabling the border

jouderianjr (approved):

src/UI/SummaryListItem.elm
```diff
@@ -0,0 +1,85 @@
...
+    let
+        colors =
+            listItemColors isSelected
+    in
+    Keyed.row
+        [ Element.width fill
+        , Element.paddingEach { top = 11, bottom = 11, left = 20, right = 12 }
+        ]
+        [ ( "label", listItemLabel renderConfig colors title caption )
+        , ( "badge", listItemBadge renderConfig colors.badge number )
```
PedroHLC, line 29: :sweat_smile:  Hey man, could we change `Int` to be a `Badge` instead, so we can color them and insert text? It will be necessary for Phase 3!!

PedroHLC (changes requested):

src/UI/Badge.elm
```diff
@@ -177,6 +178,13 @@ successDark content =
     Badge { content = content, tone = ToneSuccess, brightness = Dark }
 
 
+withTone : (String -> Badge) -> Badge -> Badge
+withTone builder badge =
```
PedroHLC, line 182: why not `withTone builder (Badge {content})`?

PedroHLC (approved):

src/UI/Badge.elm
renanpvaz: right I forgot I could do that

PR #106, opened 2020-09-23
WMS-514 | Run Percy on a separate workflow

renanpvaz:
#### :thinking: What?
Moving the Percy step to its own separate job

#### :man_shrugging: Why?
So we get the test results without having to wait for Percy

#### :pushpin: Jira Issue
[WMS-514](https://paacklogistics.atlassian.net/browse/WMS-514)

### :fire: Extra
This is the time to rethink what we want to do with this step. Another option is to run it as a whole separate workflow that runs only on master, a scheduled job or even put it in the deploy workflow.


PedroHLC (approved):
:LGTM:

miguelcoba (approved):

PR #108, opened 2020-09-25
WMS-589 | Make our login page comply to LastPass (apply to password too)

PedroHLC:
#### :thinking: What?
The same as #107 but applying to password.

#### :man_shrugging: Why?
The last PR added LastPass to the username field, but they didn't sync when auto-filling.

#### :pushpin: Jira Issue
https://paacklogistics.atlassian.net/browse/WMS-589

#### :no_good: Blocked by
Not blocked.

#### :clipboard: Pending
Nothing.

### :fire: Extra
I previously misjudged this as unnecessary, sorry for the double reviewing...

### DIFF
--- a/src/UI/TextField.elm
+++ b/src/UI/TextField.elm
@@ -676,7 +676,10 @@ attrs cfg prop opt =
         usernameAttr acu =
             case prop.content of
                 ContentUsername ->
-                    Element.usernameTag :: acu
+                    Element.nameUsername :: acu
+
+                ContentPassword _ ->
+                    Element.namePassword :: acu
 
                 _ ->
                     acu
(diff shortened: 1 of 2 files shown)


renanpvaz (approved):
I missed the other PR, but like I said before we should still use a form element for accessibility.

PedroHLC:
> I missed the other PR, but like I said before we should still use a form element for accessibility.

That's not an easy task, we would need to pipe through `Element.layout`, `Html.form`, `Element.html`.
Elm-UI wasn't made for this, `Element.layout` is supposed to be used only once.

For example, doing that in `Login.View`, messed with the Alert styling, even though they're in separate parts of code.

segmentationfaulter (approved):

jouderianjr:
> I missed the other PR, but like I said before we should still use a form element for accessibility.

Well, it seems elm-ui doesn't have a form tag intentionally ([here](https://package.elm-lang.org/packages/mdgriffith/elm-ui/latest/Element-Input)).

It seems also the owner is open to [discussion](https://github.com/mdgriffith/elm-ui/issues/102) about this subject, do you think it worth creating an issue there?

PR #110, opened 2020-09-29
WMS-605 | Add an option for horizontally laid radio buttons

segmentationfaulter:
#### :thinking: What?
Add an option for horizontally laid radio buttons

#### :man_shrugging: Why?
Need them before we can merge [this PR](https://github.com/PaackEng/wms-web/pull/174)

#### :pushpin: Jira Issue
[WMS-605](https://paacklogistics.atlassian.net/browse/WMS-605)

### DIFF
--- a/cypress/integration/visual-regression.spec.js
+++ b/cypress/integration/visual-regression.spec.js
@@ -64,9 +64,9 @@ describe('Percy Snapshots', () => {
   })
 
   describe('Radio', () => {
-    it('Creates a snapshot for vertically arranged radio buttons variation', () => {
-      cy.visit('/#Basics/Radio/Vertical')
-      cy.percySnapshot('Radio Groups')
+    it('Creates a snapshot for horizontal and vertical arrangement of radio groups', () => {
+      cy.visit('/#Basics/Radio/United')
+      cy.percySnapshot('Radio groups')
     })
   })
 
--- a/cypress/integration/visual-regression.spec.js
+++ b/cypress/integration/visual-regression.spec.js
@@ -64,9 +64,9 @@ describe('Percy Snapshots', () => {
   })
 
   describe('Radio', () => {
-    it('Creates the Radio page snapshot', () => {
-      cy.visit('/#Basics/Radio/Radio')
-      cy.percySnapshot('Radio')
+    it('Creates a snapshot for vertically arranged radio buttons variation', () => {
+      cy.visit('/#Basics/Radio/Vertical')
+      cy.percySnapshot('Radio Groups')
     })
   })
 
--- a/src/UI/Radio.elm
+++ b/src/UI/Radio.elm
@@ -4,6 +4,7 @@ module UI.Radio exposing
     , withButtons, withSelected
     , RadioWidth, withWidth, widthFull, widthRelative
     , renderElement
+    , Direction, horizontal, vertical, withDirection
     )
 
 {-| Accessible and uniform-styled implementation of a radio buttons.
@@ -79,6 +80,11 @@ type RadioWidth
     | WidthRelative
 
 
+type Direction
+    = Vertical
+    | Horizontal
+
+
 type alias Properties id msg =
     { label : String
     , message : id -> msg
@@ -89,6 +95,7 @@ type alias Options id =
     { selected : Maybe id
     , buttons : List (RadioButton id)
     , width : RadioWidth
+    , direction : Direction
     }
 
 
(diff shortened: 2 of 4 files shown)


src/UI/Radio.elm
```diff
@@ -151,6 +158,23 @@ withWidth width (RadioGroup prop opt) =
     RadioGroup prop { opt | width = width }
 
 
+{-| `Radio.withDisposition` determines wheter the radio group is laid horizontally or vertically.
```
PedroHLC, line 161: `Radio.withDisposition` determines whether the radio group's items are arranged horizontally or vertically.

src/UI/Radio.elm
```diff
@@ -151,6 +158,23 @@ withWidth width (RadioGroup prop opt) =
...
+
+    Radio.withDisposition Radio.horizontalDisposition someRadioGroup
+
+-}
+withDisposition : Disposition -> RadioGroup id msg -> RadioGroup id msg
+withDisposition disposition (RadioGroup prop opt) =
+    RadioGroup prop { opt | disposition = disposition }
+
+
+{-| The buttons will be laid horizontally
```
PedroHLC, line 171: When displaying, arrange the buttons in horizontal lines.

src/UI/Radio.elm
```diff
@@ -172,7 +196,16 @@ widthRelative =
...
-renderElement renderConfig (RadioGroup { label, message } { selected, buttons, width }) =
+renderElement renderConfig (RadioGroup { label, message } { selected, buttons, width, disposition }) =
+    let
+        layoutFunc =
+            case disposition of
+                Vertical ->
+                    Element.column
+
+                Horizontal ->
+                    Element.row
```
PedroHLC, line 207: maybe `Element.wrappedRow` would be better...

src/UI/Radio.elm
```diff
@@ -4,6 +4,7 @@ module UI.Radio exposing
     , withButtons, withSelected
     , RadioWidth, withWidth, widthFull, widthRelative
     , renderElement
+    , horizontalDisposition, withDisposition
```
jouderianjr, line 7: I would add a `verticalDisposition` as well, as IMO all the "options" should be available even the default one.

jouderianjr (changes requested):
Hey, can you add a story with this? and also add to the percy story :D

PedroHLC (approved):

miguelcoba (approved):

src/UI/Radio.elm
```diff
@@ -79,6 +80,11 @@ type RadioWidth
     | WidthRelative
 
 
+type Disposition
+    = Vertical
+    | Horizontal
+
+
```
renanpvaz, line 87: Although it's not incorrect I think the word "disposition" is more often used referring to a person's mood. I would go instead with "direction" (like the flex direction), "orientation" or "alignment".

src/UI/Radio.elm
```diff
@@ -151,6 +158,23 @@ withWidth width (RadioGroup prop opt) =
...
+withDisposition disposition (RadioGroup prop opt) =
+    RadioGroup prop { opt | disposition = disposition }
+
+
+{-| When displaying, arrange the buttons in horizontal lines.
+-}
+horizontalDisposition : Disposition
+horizontalDisposition =
+    Horizontal
+
```
renanpvaz, line 176: I'd call this just `horizontal` for short, since it's the only thing using the context in the module

src/UI/Radio.elm
```diff
@@ -89,6 +95,7 @@ type alias Options id =
     { selected : Maybe id
```
renanpvaz, line 95: no change needed here, but radios mostly used with a default value already selected, and since you can't deselect it we could improve the API by getting rid of the maybe here

jouderianjr: That's true, but we would have to change the whole API in fact, because then we should not be able to create a group with an empty options list

src/UI/Radio.elm
```diff
@@ -163,18 +163,25 @@ withWidth width (RadioGroup prop opt) =
     Radio.withDisposition Radio.horizontalDisposition someRadioGroup
```
PedroHLC, line 163: :smile:  remind to also update the example in docs

cypress/integration/visual-regression.spec.js
```diff
@@ -64,9 +64,9 @@ describe('Percy Snapshots', () => {
   })
 
   describe('Radio', () => {
```
jouderianjr, line 66: Hey, it's better to have one story containing both examples, like we have for buttons

jouderianjr (changes requested):

jouderianjr (approved):

PR #111, opened 2020-09-30
Add UI.Utils.Action

PedroHLC:
#### :thinking: What?
Add a configuration for components that defines their own actions, like `ListView` and `StackHeader`

#### :man_shrugging: Why?
Removes a "hole" in the code.

#### :pushpin: Jira Issue
Not tracked.

#### :no_good: Blocked by
Not blocked, based on master.

#### :clipboard: Pending
- [ ] Documentation.


jouderianjr (approved):

src/UI/Utils/Action.elm
```diff
@@ -0,0 +1,60 @@
...
+    , action =
+        DialogMsg.OpenElementCreation
+            |> Msg.ForDialog
+            |> Action.DispatchMsg
+    }
+
+-}
+type alias Config msg =
+    { action : Action msg
+    , icon : String -> Icon
```
renanpvaz, line 32: will it always have an icon?

src/UI/Utils/Action.elm
```diff
@@ -0,0 +1,60 @@
+module UI.Utils.Action exposing
```
renanpvaz, line 1: feels a little weird having this under utils. What about something like `UI.Config.Action`, `UI.Data.Action` or even just `UI.Action`?

src/UI/Internal/Clickable.elm
```diff
@@ -0,0 +1,45 @@
+module UI.Internal.Clickable exposing (actionIcon, actionWrapElement, msgWrapElement)
+
+import Element exposing (Attribute, Element)
+import UI.Icon as Icon
+import UI.Link as Link
+import UI.RenderConfig exposing (RenderConfig)
+import UI.Utils.ARIA as ARIA
+import UI.Utils.Action as Action exposing (Action(..))
```
renanpvaz, line 8: looking at it again it seems like it's all about the icons. Doesn't it make sense to merge UI.Utils.Action and UI.Internal.Clickable under a single `UI.Icon.Action`?

PedroHLC:
@renanpvaz 
I don't see the Actions as full components. They're pretty much a utility for describing small (clickable) parts of other components.

While `UI.Internal.Clickable` may not have been the greatest idea, but It's important to remember they're not "how Actions should render", and instead just helpers for you to achieve that!

And they're not 100% icons related, like on ListView's action bar, where it's an entire clickable block with text beside icons.

src/UI/Utils/Action.elm
PedroHLC: This one is a great question. Previously I've asked @jouderianjr  the same, and his answer was: it doesn't make much sense to not having it.
While my opinion is: having a clickable text, or any other non-icon component, will just add enough abstraction that would make more sense to use elm-ui directly.

jouderianjr: My statement is a bit out of context here :o hahaha. But when I said that, I was talking more specifically to the StackChild, not exactly in general.

But I'm not sure if it makes sense to support any other thing here 🤔  we don't have right now an example of that because in the end this Module is used by other modules and not in elm-ui directly.

jouderianjr:
That's a good discussion, I'm more or less with @PedroHLC on that, But let's understand the problem first:

Our problem here is because we have some custom components that have clickable areas but it's not a button, then we can reuse the whole "what should a clickable container looks like" in that different components instead of implement over and over.

The action is because when we have those clickable places, it could trigger a message or could be a link, then it would make sense to unify this "interface".

I'm just wondering here if makes sense to have the `icon` field here or not 🤔 , if we want to be really generic, we could have ActionConfig without any UI-related field and then we can specify the UI in the component itself.

Something like:

```elm
withActionBar :
    (Action.Config msg, Icon)
    -> ListView object msg
    -> ListView object msg
```

renanpvaz (approved):

PR #116, opened 2020-10-19
Add squared button

renanpvaz:
#### :thinking: What?
Add a new button variation that is the primary theme but with squared borders

#### :man_shrugging: Why?
Because it is needed for the rule creation form in the BRMS and it's going to be added to the Design System according to Sufiyan.

#### :pushpin: Jira Issue
[BRMS-74](https://paacklogistics.atlassian.net/browse/BRMS-74)

### DIFF
--- a/showcase/src/Buttons/Stories.elm
+++ b/showcase/src/Buttons/Stories.elm
@@ -42,6 +42,7 @@ stories renderConfig =
         , dangerStory renderConfig
         , lightStory renderConfig
         , clearStory renderConfig
+        , squaredStory renderConfig
         , linkStory renderConfig
         , fullWidthStory renderConfig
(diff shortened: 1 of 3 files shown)


PedroHLC (approved):

jouderianjr:
I'm not sure if we can assume that the "squared button" will be always a primary button, what happens if this we have a Warning squared button in the future? it will be complicated to add it.

What if we create a new variation of `withWidth`? something like `widthFullSquared`? this way we support all the "color" variations. Does that make sense?

miguelcoba (approved):

PedroHLC:
stalled and (according to @renanpvaz) also waiting a button refactor

PR #118, opened 2020-10-28
WMS-660 | Add "Unfiltrable" column to Tables

PedroHLC:
(Why is this a Draft? Cause I don't know if I'm gonna use it yet!)


PedroHLC:
We didn't have to use this. So closing it...

PR #120, opened 2020-11-04
WMS-676 | Move cellLink to UI.Table

PedroHLC:
#### :thinking: What?
* Move `cellLink` to UI.Table

#### :man_shrugging: Why?
* This code was in the app and may be reused.

#### :pushpin: Jira Issue
[WMS-676](https://paacklogistics.atlassian.net/browse/WMS-676)

#### :no_good: Blocked by
Not blocked.

#### :clipboard: Pending
Nothing.

### DIFF
--- a/src/UI/Internal/Tables/View.elm
+++ b/src/UI/Internal/Tables/View.elm
@@ -3,6 +3,7 @@ module UI.Internal.Tables.View exposing (..)
 import Element exposing (Attribute, Element, fill, fillPortion, minimum, px, shrink)
 import Element.Background as Background
 import Element.Border as Border
+import Element.Font as Font
 import Element.Keyed as Keyed
 import UI.Button as Button
 import UI.Internal.NArray as NArray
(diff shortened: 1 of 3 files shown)


src/UI/Internal/Tables/View.elm
```diff
@@ -47,6 +52,26 @@ simpleText renderConfig text =
             ]
 
 
+linkedText : RenderConfig -> Link -> Text -> Element msg
+linkedText renderConfig link text =
+    text
+        |> Text.withOverflow ellipsizeWithTooltip
+        |> Text.withColor (Palette.color tonePrimary brightnessMiddle)
+        |> Text.renderElement renderConfig
```
renanpvaz, line 60: what if we need links with the overflow behavior outside the table? We'd have to copy it over, so maybe we should create a standalone link component outside of the button. Just something to keep in mind

renanpvaz (approved):

src/UI/Internal/Tables/View.elm
PedroHLC: I find it interesting, could you propose the API for that? Also, CC @jouderianjr

jouderianjr: yeah, taking in mind the overflow, it makes sense to have it also as a component

jouderianjr (approved):

src/UI/Internal/Tables/View.elm
jouderianjr: We can do it after the BF thing, can you create a card for that @PedroHLC ?

PR #121, opened 2020-11-09
WMS-686 | UI.TextField changes for user creation

PedroHLC:
#### :thinking: What?
* Add `TextField.toStaticIf`, so any field can become static given some condition;
* Show labels when static;
* Hide passwords when static;
* Automatically use placeholders for passwords

#### :man_shrugging: Why?
* To keep UI logic in UI.

#### :pushpin: Jira Issue
[WMS-686](https://paacklogistics.atlassian.net/browse/WMS-686)

#### :no_good: Blocked by
Not blocked.

#### :clipboard: Pending
Nothing.

### :fire: Extra
a gif to make you smile

### DIFF
--- a/src/UI/TextField.elm
+++ b/src/UI/TextField.elm
@@ -4,7 +4,7 @@ module UI.TextField exposing
     , newPassword, currentPassword, setPasswordVisible
     , username, email
     , search
-    , static
+    , static, toStaticIf
     , withSize
     , TextFieldWidth, withWidth, widthFull, widthRelative
     , setLabelVisible, withPlaceholder, withIcon
@@ -59,7 +59,7 @@ Different from [`Element.Input`](/packages/mdgriffith/elm-ui/latest/Element-Inpu
 
 ## Static
 
-@docs static
+@docs static, toStaticIf
 
 
 # Size
@@ -159,6 +159,29 @@ type TextFieldWidth
 -- Options
 
 
+{-| Irreversibly converts the input to static field. Given True to a condition.
+
+    TextField.newPassword Msg.OnPasswordChanged
+        "New password"
+        model.value
+        |> TextField.toStaticIf
+            model.hasSentForm
+
+-}
+toStaticIf : Bool -> TextField msg -> TextField msg
+toStaticIf condition =
+    if condition then
+        \(TextField prop opt) ->
+            TextField
+                { prop
+                    | changeable = Nothing
+                }
+                opt
+
+    else
+        identity
+
+
 {-| Listen to focus events, add tab-indexing and enforce focus state.
 
     TextField.withFocus
@@ -197,6 +220,8 @@ withIcon icon (TextField prop opt) =
 {-| Place-holds the text field with text.
 
     TextField.withPlaceholder "Enter your personal email" someTextField
+
+**NOTE**: When creating password fields, the placeholder is "●●●●●●●●" by default.
 
 -}
 withPlaceholder : String -> TextField msg -> TextField msg
@@ -363,6 +388,8 @@ username onChange label currentValue =
         "New password"
         model.value
 
+**NOTE**: Uses [`withPlaceholder "●●●●●●●●"`](#withPlaceholder) as default.
+
 -}
 newPassword : (String -> msg) -> String -> String -> TextField msg
 newPassword onChange label currentValue =
@@ -370,6 +397,7 @@ newPassword onChange label currentValue =
         label
         currentValue
         forNewPassword
+        |> withPlaceholder "●●●●●●●●"
 
 
 {-| Wrapper around [`Element.Input.currentPassword` ](/packages/mdgriffith/elm-ui/latest/Element-Input#currentPassword).
@@ -378,6 +406,8 @@ newPassword onChange label currentValue =
         "Current password"
         model.value
 
+**NOTE**: Uses [`withPlaceholder "●●●●●●●●"`](#withPlaceholder) as default.
+
 -}
 currentPassword : (String -> msg) -> String -> String -> TextField msg
 currentPassword onChange label currentValue =
@@ -385,6 +415,7 @@ currentPassword onChange label currentValue =
         label
         currentValue
         forCurrentPassword
+        |> withPlaceholder "●●●●●●●●"
 
 
 {-| Wrapper around [`Element.Input.email` ](/packages/mdgriffith/elm-ui/latest/Element-Input#email).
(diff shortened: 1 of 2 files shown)


src/UI/TextField.elm
```diff
@@ -159,6 +159,29 @@ type TextFieldWidth
...
+{-| Irreversibly converts the input to static field. Given True to a condition.
+
+    TextField.newPassword Msg.OnPasswordChanged
+        "New password"
+        model.value
+        |> TextField.toStaticIf
+            model.hasSentForm
+
+-}
+toStaticIf : Bool -> TextField msg -> TextField msg
```
jouderianjr, line 171: I would keep using `with*`, we do have something similar in the UI.Button, I would call it `withStaticIf`.

That being said, I would also rename the term used here, I'd prefer disabled instead of static, I think `disabled` is widely used and it's more natural in my opinion in the frontend world.

src/UI/TextField.elm
```diff
@@ -460,59 +491,99 @@ The result of this function is a ready-to-insert Elm UI's Element.
...
                 |> Element.el elAttrs
 
-        whenPassword msg { isVisible, isCurrent } =
-            if isCurrent then
-                inputPasswordOptions cfg msg prop opt isVisible
-                    |> Input.currentPassword elAttrs
+        inputElement =
+            case prop.content of
+                ContentPassword _ ->
+                    "●"
```
jouderianjr, line 561: Why not asterisks? I know we used this character in the wms-web, but I don't think makes sense to change the format. I would suggest using asterisks.

src/UI/TextField.elm
```diff
@@ -378,13 +406,16 @@ newPassword onChange label currentValue =
...
+**NOTE**: Uses [`withPlaceholder "●●●●●●●●"`](#withPlaceholder) as default.
+
 -}
 currentPassword : (String -> msg) -> String -> String -> TextField msg
 currentPassword onChange label currentValue =
     input onChange
         label
         currentValue
         forCurrentPassword
+        |> withPlaceholder "●●●●●●●●"
```
jouderianjr, line 418: why force this? I didn't get it

jouderianjr (changes requested):
It's missing our lovely stories :D

src/UI/TextField.elm
PedroHLC: Because Firefox and WebKit-based browsers use filled-circles tas password security character

src/UI/TextField.elm
PedroHLC: :sweat_smile:  It's the "best default", as it's what we'll use everywhere...

src/UI/TextField.elm
PedroHLC: The thing about `to` is: This is an irreversible change to the TextField.

> I think disabled is widely used and it's more natural in my opinion in the frontend world.

Static fields can't be "reenabled". They use a different construction process and even aren't real input fields. They're common elements mimicing the input appearence.

Another solution would be to eradicate the current way of creating a static field and make the message an optional/maybe field for all constructors.

PedroHLC:
> It's missing our lovely stories :D
Indeed, I'll start it right now!

src/UI/TextField.elm
```diff
@@ -298,14 +323,14 @@ widthFull =
...
+{-| The field will have the width you specify in pixels.
 
-**NOTE**: Default behaviour.
+**NOTE**: Default behaviour with 260px.
 
 -}
-widthRelative : TextFieldWidth
-widthRelative =
-    WidthRelative
+widthPx : Int -> TextFieldWidth
```
PedroHLC, line 331: There is now a collateral change in this PR.

The problem with `widthRelative` was that it wasn't relative at all. It was defaulting to a browser's value, inherited through the "label" field.

src/UI/TextField.elm
jouderianjr: I didnt get why it is irreversible, this function receives a boolean, AFAIK if the boolean it's true the field starts to work again right? in the create a user story, if the backend returns an error, we enabled the field again right?

PedroHLC: > if the boolean it's true the field starts to work again right?

Nope, this flow would happen:
* construct (with `currentPassword`, `username` , etc) -> TextField with _Msg_ and with _Type_ -> `withStaticIf True` -> TextField with _Type_ and without _Msg_
* construct (with `static`) -> TextField without _Msg_ and without _Type_ -> `withStaticIf False` -> **missing** _Msg_ and _Type_

Reminding you that a disabled/static doesn't have a message cause we don't want to creat a "Msg.NoOp" for a static field..

A better API would have all the constructors with `Maybe Msg`, where being Nothing means it's disabled. This way we could create passwords and multilines "disabled" fields.

(I'm not enforcing either of this two solutions, I'm really open to discussion about this)

src/UI/TextField.elm
```diff
@@ -785,3 +774,88 @@ textFieldPadding size =
...
+                |> Text.subtitle2
+                |> Text.renderElement renderConfig
+                |> Element.el elAttrs
+
+        inputElement =
+            case prop.content of
+                ContentPassword _ ->
+                    "●"
+                        |> String.repeat (String.length prop.currentValue)
+                        |> genericInputElement
```
renanpvaz, line 840: Why not just use the password input type?

PedroHLC: elm-ui docs recommend this: https://package.elm-lang.org/packages/mdgriffith/elm-ui/latest/Element-Input#disabling-inputs

PedroHLC: > For other inputs such as Input.text, consider simply rendering it in a normal paragraph or el if it's not editable.
Alternatively, see if it's reasonable to not display an input if you'd normally disable it. Is there an option where it's only visible when it's editable?

renanpvaz: Hmm not sure this applies to the input case, the discussion linked in the documentation is about a button that allows the user to progress. After reading it through I actually believe now that our "fake" disabled behavior could be bad, because there's nothing but the visual cue to indicate that it is disabled, so it could be  interpreted as not working

src/UI/TextField.elm
renanpvaz: I think we might be stepping our own toes here. I don't see reason not to use the "real" disabled behavior. Turning it on and off would not longer be a problem in that case and it would be much simpler too.

renanpvaz: Moreover, I think we should lean into the language we're familiar with when it makes sense. I didn't immediately get what the whole "static" thing was about, so calling it either disabled or read only would be better in that sense

src/UI/TextField.elm
PedroHLC: I agree with you, it doesn't seem that having a field that's "always disabled" will be useful.

As this is impossible in elm-ui, by now let's go with killing the `static` initializer, but reusing it in a `withDisabledIf`

PR #122, opened 2020-11-10
BRMS-37 | Add move and trash icons

renanpvaz:
#### :thinking: What?
Two missing icons

#### :man_shrugging: Why?
Because I need them and they were in the assets already

#### :pushpin: Jira Issue
[BRMS-37](https://paacklogistics.atlassian.net/browse/BRMS-37)

### DIFF
--- a/src/UI/Icon.elm
+++ b/src/UI/Icon.elm
@@ -4,7 +4,7 @@ module UI.Icon exposing
     , add, check, close, collapse, download, edit, eventLog, expand, filter
     , groups, logout, move, nextContent, notifications, paackSpaces, packages
     , previousContent, print, remove, sandwichMenu, search, searchSpace
-    , seeMore, toggle, toggleDown, toggleUp, trash, warning
+    , seeMore, toggle, toggleDown, toggleUp, delete, warning
     , getHint
     , withColor
     , withSize, withCustomSize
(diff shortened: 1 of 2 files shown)


src/UI/Icon.elm
```diff
@@ -443,6 +445,26 @@ remove hint =
     Icon (Properties hint Remove) defaultOptions
 
 
+{-| A trash can with an "x" on it.
+
+    Icon.trash "Delete item"
+
+-}
+trash : String -> Icon
```
PedroHLC, line 453: Icons are named after their action, so we don't misuse it to designate two different actions.
In this case, I prefer it named `Delete`.

PedroHLC (changes requested):

PedroHLC (approved):
:LGTM:

jouderianjr (approved):

src/UI/Icon.elm
```diff
@@ -443,6 +445,26 @@ remove hint =
     Icon (Properties hint Remove) defaultOptions
 
 
+{-| A trash can with an "x" on it.
+
+    Icon.delete "Delete item"
+
+-}
+delete : String -> Icon
```
jouderianjr, line 453: check this Item because I cant see it in the showcase

jouderianjr (changes requested):

src/UI/Icon.elm
renanpvaz: oops find n replace gone wrong

jouderianjr (approved):

PR #125, opened 2020-11-30
Fix incompatible paddings between text and links in Tables

PedroHLC:
#### :thinking: What?
Share the same attributes for text and links.

#### :man_shrugging: Why?
The padding was different.

#### :pushpin: Jira Issue
Related to WMS-726.

#### :no_good: Blocked by
Not blocked, based on master.

#### :clipboard: Pending
Nothing.


jouderianjr (approved):
As we are here, I would add a Link to the percy example table :D :D

renanpvaz (approved):

PR #127, opened 2020-12-01
RFC-005 | Palette color shorthand

renanpvaz:
#### :thinking: What?
Add one shorthand per color variant.
Also had to fix ambiguous imports

#### :man_shrugging: Why?
To reduce verbosity when using colors

### DIFF
--- a/src/UI/Internal/Tables/View.elm
+++ b/src/UI/Internal/Tables/View.elm
@@ -7,7 +7,7 @@ import Element.Font as Font
 import Element.Keyed as Keyed
 import UI.Button as Button
 import UI.Internal.NArray as NArray
-import UI.Internal.Palette as Palette
+import UI.Internal.Palette as Colors
 import UI.Internal.Primitives as Primitives
 import UI.Internal.Tables.Common exposing (..)
 import UI.Internal.Utils.Element as InternalElement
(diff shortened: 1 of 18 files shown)


src/UI/Internal/Tables/View.elm
```diff
@@ -7,7 +7,7 @@ import Element.Font as Font
 import Element.Keyed as Keyed
 import UI.Button as Button
 import UI.Internal.NArray as NArray
-import UI.Internal.Palette as Palette
+import UI.Internal.Palette as Colors
```
jouderianjr, line 10: I think called this using a different name it's even more confusing. I personally don't like this "importing overlap" we have in the project  (Internal.Palette and Paletted having the same alias), I would prefer tag something like `IPalette` to show explicit is an internal module. But changing the name IMO it's a bit misleading.

(we can discuss and decide this as a standard approach how to use these "overlapped imports")

cc @PedroHLC, he is the one that likes this current approach.

PedroHLC: :sweat_smile: Thinking about it, "Colors" is the most appropriate name for the Internal module, as it really doesn't know at all about our Palette type.

As this is probably the last addition to the V1 of paack-ui, I think we could go with it...

PedroHLC (approved):

src/UI/Internal/Tables/View.elm
jouderianjr: Then I would prefer rename `Ui.Internal.Palette` to `Ui.Internal.Colors`

jouderianjr (approved):

PR #129, opened 2020-12-04
WMS-715 | Add v2 of dialog as a POC

segmentationfaulter:
#### :thinking: What?
Adds another Dialog component as discussed in this [RFC](https://paacklogistics.atlassian.net/wiki/spaces/EN/pages/2089025545/New+Dialog+API),

#### :man_shrugging: Why?
Check out the RFC

#### :pushpin: Jira Issue
[WMS-715](https://paacklogistics.atlassian.net/browse/WMS-715)

### :fire: Extra
I have made icon in the title as mandatory to create this version of the dialog and removed close message as an argument since this version of the dialog is not going to have a cross button to close the dialog nor it will have the overlay. Have a look at [this conversation](https://www.figma.com/file/DTyoSveNksHpZCdU8a7BNs?node-id=369:29#50879134) regarding the title in the icon and removal of the cross button.

### DIFF
--- a/showcase/src/Dialog.elm
+++ b/showcase/src/Dialog.elm
@@ -1,0 +1,63 @@
+module Dialog exposing (stories)
+
+import Element
+import Msg
+import PluginOptions exposing (defaultWithoutMenu)
+import UI.Button as Button exposing (Button)
+import UI.Icon as Icon
+import UI.Internal.Palette exposing (overlayBackground)
+import UI.RenderConfig exposing (RenderConfig)
+import UI.Text as Text
+import UI.V2.Dialog as Dialog
+import UIExplorer exposing (storiesOf)
+import Utils exposing (ExplorerStory, ExplorerUI, iconsSvgSprite, prettifyElmCode, story)
+
+
+stories : RenderConfig -> ExplorerUI
+stories cfg =
+    storiesOf
+        "Dialog"
+        [ dialog cfg
+        ]
+
+
+dialog : RenderConfig -> ExplorerStory
+dialog cfg =
+    story
+        ( "Dialog"
+        , Element.column
+            [ Element.centerX
+            , Element.centerY
+            , Element.padding 32
+            , overlayBackground
+            ]
+            [ iconsSvgSprite
+            , Dialog.Dialog
+                { title = "Dialog Header"
+                , icon = Icon.warning "Warning dialog"
+                }
+                { buttons = buttons
+                , body = "I am body of the dialog" |> Text.body2 |> Text.renderElement cfg
+                }
+                |> Dialog.renderElement cfg
+            ]
+        , { defaultWithoutMenu | code = code }
+        )
+
+
+buttons : List (Button Msg.Msg)
+buttons =
+    [ Button.fromLabel "Ok" |> Button.cmd Msg.NoOp Button.primary
+    , Button.fromLabel "Ok" |> Button.cmd Msg.NoOp Button.primary
+    , Button.fromLabel "Ok" |> Button.cmd Msg.NoOp Button.primary
+    ]
+
+
+code : String
+code =
+    prettifyElmCode """
+dialogV2 "Dialog title" Icon.warning
+    |> Dialog.withBody ("Dialog body text" |> Text.body2 |> Text.renderElement cfg)
+    |> Dialog.withButtons
+        (Button.fromLabel "Ok" |> Button.cmd NoOp Button.primary)
+"""
--- a/showcase/src/Dialog.elm
+++ b/showcase/src/Dialog.elm
@@ -6,10 +6,10 @@ import PluginOptions exposing (defaultWithoutMenu)
 import UI.Button as Button exposing (Button)
 import UI.Icon as Icon
 import UI.Internal.Colors exposing (overlayBackground)
+import UI.Internal.DialogV2 exposing (dialogViewV2)
 import UI.RenderConfig exposing (RenderConfig)
 import UI.Text as Text
 import UI.V2.Dialog as Dialog
-import UI.Internal.DialogV2 exposing(dialogViewV2)
 import UIExplorer exposing (storiesOf)
 import Utils exposing (ExplorerStory, ExplorerUI, iconsSvgSprite, prettifyElmCode, story)
 
--- a/showcase/src/Dialog.elm
+++ b/showcase/src/Dialog.elm
@@ -9,6 +9,7 @@ import UI.Internal.Colors exposing (overlayBackground)
 import UI.RenderConfig exposing (RenderConfig)
 import UI.Text as Text
 import UI.V2.Dialog as Dialog
+import UI.Internal.DialogV2 exposing(dialogViewV2)
 import UIExplorer exposing (storiesOf)
 import Utils exposing (ExplorerStory, ExplorerUI, iconsSvgSprite, prettifyElmCode, story)
 
@@ -28,7 +29,8 @@ dialog cfg =
         , Element.column
             [ Element.centerX
             , Element.centerY
-            , Element.padding 100
+            , Element.padding 20
+            , Element.height <| Element.px 500
             , overlayBackground
             ]
             [ iconsSvgSprite
@@ -38,7 +40,7 @@ dialog cfg =
                 |> Dialog.withBody
                     ("Dialog body text" |> Text.body2 |> Text.renderElement cfg)
                 |> Dialog.withButtons buttons
-                |> Dialog.renderElement cfg
+                |> dialogViewV2 cfg
             ]
         , { defaultWithoutMenu | code = code }
         )
--- a/showcase/src/Dialog.elm
+++ b/showcase/src/Dialog.elm
@@ -28,17 +28,16 @@ dialog cfg =
         , Element.column
             [ Element.centerX
             , Element.centerY
-            , Element.padding 32
+            , Element.padding 100
             , overlayBackground
             ]
             [ iconsSvgSprite
-            , Dialog.Dialog
-                { title = "Dialog Header"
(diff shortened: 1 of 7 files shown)


miguelcoba (approved):

showcase/src/Dialog.elm
```diff
@@ -0,0 +1,63 @@
+module Dialog exposing (stories)
```
PedroHLC, line 1: Please, add it to Percy too.

showcase/src/Dialog.elm
```diff
@@ -0,0 +1,63 @@
...
+    story
+        ( "Dialog"
+        , Element.column
+            [ Element.centerX
+            , Element.centerY
+            , Element.padding 32
+            , overlayBackground
+            ]
+            [ iconsSvgSprite
+            , Dialog.Dialog
```
PedroHLC, line 35: I think it's better to use the same code as the example in `code`.

src/UI/NavigationContainer.elm
```diff
@@ -145,7 +147,8 @@ type alias Content msg =
 See [`Nav.dialog`](#dialog) to see how to create a dialog.
 -}
 type Dialog msg
-    = Dialog (Dialog.Dialog msg)
+    = Dialog1 (Dialog1.Dialog msg)
```
PedroHLC, line 150: You don't need this. It could be:
```
type alias Dialog msg =
   (Dialog2.Dialog ms)
```
The dialog v1 can be described as a v2.

src/UI/NavigationContainer.elm
```diff
@@ -434,13 +437,37 @@ Clicking on the black layer also activates the closing message.
 -}
 dialog : String -> msg -> Element msg -> Dialog msg
 dialog title onClose body =
-    Dialog
+    Dialog1
```
PedroHLC, line 440: Here it would become:
```
        Dialog2.Dialog
            { title = title
            , icon = Nothing
            }
            { buttons = []
            , body = body
            }
````

File: src/UI/NavigationContainer.elm [hunk 5]
```diff
@@ -434,13 +437,37 @@ Clicking on the black layer also activates the closing message.
...
 
 
+{-|
+
+    `Nav.dialogV2` constructs a [`Nav.Dialog`](#Dialog) from a title and icon.
+    This variant of the dialog does not require you to specify body at the time
+    of creation and you can specify it alongside dialog buttons as an option
+    like this:
+
+    Nav.dialog2 "An example dialog" Icon.warning
```
PedroHLC, line 454: Okay, now I see, this part is wrong.

The idea is to have this dialog a totally separated dialog. So it would do:
```
import UI.V2.Dialog as Dialog
-- ...

container : AppConfig -> Nav.Container msg
container { locale, renderConfig } =
    { title = locale.t "pages.404.title"
    , content = Nav.contentSingle <| view locale renderConfig
    , dialog = Nothing -- Deprecated: use `withDialog` instead
    , hasMenu = False
    }
    |> Nav.containerWithDialog
            (Dialog.dialog "An example dialog"
                     (Element.text "Dialog body")
                     |> Dialog.withIcon Icon.warning
                     |> Dialog.withButtons [submitButton, cancelButton]
             )
```

And when rendering it in showcase, it would be something like:
```elm
            Dialog.dialog "An example dialog"
                     (Element.text "Dialog body")
                     |> Dialog.withIcon Icon.warning
                     |> Dialog.withButtons [submitButton, cancelButton]
                     |> Dialog.renderElement renderConfig
```

I'm adding this back into the RFC...

PedroHLC (changes requested):

src/UI/V2/Dialog.elm
```diff
@@ -0,0 +1,119 @@
...
+desktopDialogView cfg (Dialog { title, icon } { body, buttons }) =
+    Element.column
+        [ Element.width shrink
+        , Element.centerY
+        , Element.centerX
+        , mainBackground
+        , Element.padding 32
+        , Border.rounded 6
+        ]
+        [ desktopHeaderRow cfg title icon
```
PedroHLC, line 72: I think it's good to keep using `Element.above`.
The way you've done here, we can't set the width of the dialog from within the page (by using `Element.above` the dialog width will be the same as the body, and the header will match them Otherwise, if the header is longer than the body, it will be the one setting the dialog's width).

src/UI/V2/Dialog.elm
```diff
@@ -0,0 +1,119 @@
...
+
+dialogMap : (a -> b) -> Dialog a -> Dialog b
+dialogMap applier (Dialog { title, icon } { body, buttons }) =
+    Dialog
+        { title = title
+        , icon = icon
+        }
+        { body = Element.map applier body
+        , buttons = List.map (Button.map applier) buttons
+        }
```
renanpvaz, line 44: I'm failing to see when this function would be useful. When would we need to apply the same function to both the body and the buttons? Perhaps it makes sense to map the whole options record so that the user has the option to keep one property untouched

renanpvaz (approved):

src/UI/NavigationContainer.elm
jouderianjr: I do prefer @segmentationfaulter version, we don't want to replace the old version, we want to support both. so the old version should be available.

src/UI/NavigationContainer.elm
jouderianjr: I didn't get why have this, Nav.container already have a way to receive a dialog, having `containerWithDialog` makes things inconsistent in my opinion as you will be able to add dialog in the record and then have the with* function.

Again, we want retrocompatibility, even though we set the old version as deprecated. 

So, IMO makese sense to have:

```elm
    { title = locale.t "pages.404.title"
    , content = Nav.contentSingle <| view locale renderConfig
    , dialog = Just (Nav.dialog blablabla) | Just (Nav.dialogV2 myDialog)
    , hasMenu = False
    }
```

src/UI/NavigationContainer.elm
```diff
@@ -434,13 +437,37 @@ Clicking on the black layer also activates the closing message.
...
+        |> withBody (Element.text "Dialog body")
+        |> withButtons [submitButton, cancleButton]
+
+-}
+dialogV2 : String -> Icon -> Dialog msg
```
jouderianjr, line 459: I think it makes more sense to receive `UI.V2.Dialog.Dialog` instead of `String -> Icon`, this way the user will be able to set the dialog the way they want

src/UI/NavigationContainer.elm
jouderianjr: I would add to the docs comment saying this deprecated we must use the other one

src/UI/V2/Dialog.elm
```diff
@@ -0,0 +1,119 @@
...
+                , bottom = 0
+                , left = 0
+                , right =
+                    0
+                }
+            ]
+            body
+        , Element.row
+            [ Element.spacing 16
+            , Element.paddingEach { top = 22, left = 0, right = 0, bottom = 0 }
```
jouderianjr, line 85: I know that's exactly the same number in figma, but I would use 24 instead, Sufiyan already told us that he is using 8dot grid, so all the spacing must to be 8 multipler

src/UI/NavigationContainer.elm
PedroHLC: Yep, you're right, I've edited my comment here and will do the same in the RFC...

PedroHLC:
A note before merging:

PedroHLC:
A note before merging:
* [ ] Add this module to the exposed modules through `elm.json`

src/UI/NavigationContainer.elm
segmentationfaulter: If we make it receive `UI.V2.Dialog.Dialog`, then having the helpers like `UI.V2.withBody` and `UI.V2.withButtons` are needless I think. Perhaps you meant that `dialogV2` should receive `UI.V2.Properties`?

PedroHLC: @jouderianjr notice that the dialogV2 is not some `UI.V2.Dialog -> Nav.Container -> Nav.Container`, it's the constructor that should actually been in `UI.V2.Dialog.dialog`.

src/UI/V2/Dialog.elm
```diff
@@ -0,0 +1,205 @@
...
+
+
+
+-- Component handling
+
+
+{-| Transforms the message produced by the component.
+-}
+dialogMap : (a -> b) -> Dialog a -> Dialog b
+dialogMap applier (Dialog { title, icon, close } { body, buttons }) =
```
PedroHLC, line 90: I think it would be called just `map` and not  `dialoMap` :smile:

PedroHLC (changes requested):
Just this small change and then it's approved :smile:

src/UI/V2/Dialog.elm
```diff
@@ -0,0 +1,132 @@
...
+{-| Allows you to specify whether clicking on overlay closes the dialog or not.
+By default, clicking on overlay does nothing, that is because an unintentional
+click can cause loss of information if a dialog is open which is expecting an
+input from the user and user has already provided it partially. It is
+recommended to provide an explicit cancel/close button for such cases. However
+this helper can be used to recieve clicks on overlay to close the dialog where
+loss of information is not likely to happen.
+-}
+withCloseOnOverlayClick : Bool -> Dialog msg -> Dialog msg
+withCloseOnOverlayClick shouldClose (Dialog props options) =
```
PedroHLC, line 131: I would not publish this one yet.
Adding new exposed functions bumps minor versions, but removing them are considered api-breaking and bumps major versions.

PedroHLC (approved):

jouderianjr (approved):

PR #130, opened 2020-12-10
Allow scrolling to element in list

PedroHLC:
#### :thinking: What?
* Add id to elements in ListView when  `ListView.withDomIds "somePrefix"` is used

#### :man_shrugging: Why?
* Required for https://github.com/PaackEng/wms-web/pull/242

#### :pushpin: Jira Issue
Not tracked.

#### :no_good: Blocked by
Not blocked.

#### :clipboard: Pending
Nothing

### DIFF
--- a/src/UI/Internal/Utils/Element.elm
+++ b/src/UI/Internal/Utils/Element.elm
@@ -33,6 +33,13 @@ title value =
         |> Element.htmlAttribute
 
 
+id : String -> Attribute msg
+id value =
+    value
+        |> HtmlAttrs.id
+        |> Element.htmlAttribute
+
+
 overlayBackground : msg -> Element msg
 overlayBackground onClickMsg =
     Element.el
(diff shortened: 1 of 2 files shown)


renanpvaz (approved):

src/UI/Internal/Utils/Element.elm
```diff
@@ -33,6 +33,13 @@ title value =
         |> Element.htmlAttribute
 
 
+id : String -> Attribute msg
```
jouderianjr, line 36: Mental note: Maybe we would want to have  this in the future as an exposed function

src/UI/ListView.elm
```diff
@@ -351,6 +353,25 @@ withSelectStyle style (SelectList prop opt) =
...
+
+    < ... id="wineBrands">
+        <... id="wineBrands-key1"></...>
+        <... id="wineBrands-key2"></...>
+        <... id="wineBrands-key3"></...>
+    </...>
+
+-}
+withDomIds : String -> ListView object msg -> ListView object msg
+withDomIds containerId (SelectList prop opt) =
```
jouderianjr, line 371: was I mentioned on slack, I would prefer to have `withDomId: String -> ListView -> ListView` and this only changes the parent id, I would use the `toKey` to set the children ids( in fact this maybe should be called toId :D ).

jouderianjr (approved):

src/UI/ListView.elm
jouderianjr: Adding that we could `withDomId` as a standard for other components as well

PR #142, opened 2020-12-21
WMS-671 | Move to elm-review

segmentationfaulter:
#### :thinking: What?
Replace elm-analyse with elm-review

#### :man_shrugging: Why?
Because of [this bug](https://github.com/stil4m/elm-analyse/issues/198) which makes elm-analyse silently fail if there is a numeric character in an exposed module.

#### :pushpin: Jira Issue
[WMS-671](https://paacklogistics.atlassian.net/browse/WMS-671)

### :fire: Extra
I translated from elm-analyse rule by rule, and below are the rules which I have not included:

- [UnnecessaryParens](https://stil4m.github.io/elm-analyse/#/messages/UnnecessaryParens): Since elm-format removes these unnecessary parenthesis, we don't need this rule.
- [UnnecessaryPortModule](https://stil4m.github.io/elm-analyse/#/messages/UnnecessaryPortModule): Ports cannot be present in packages that's why we don't need it here, moreover the use-case is covered by elm-compiler itself.
- [DuplicateImport](https://stil4m.github.io/elm-analyse/#/messages/DuplicateImport): Again elm-format removes all duplicate imports, rendering the rule redundant.
- [FileLoadFailed](https://stil4m.github.io/elm-analyse/#/messages/FileLoadFailed): Couldn't understand the purpose of this rule, do we really need it?
- [MapNothingToNothing](https://stil4m.github.io/elm-analyse/#/messages/MapNothingToNothing): Couldn't find a package for it for elm-review, if we really need it, we can write a custom rule.


src/UI/Button.elm
```diff
@@ -507,15 +506,15 @@ hyperlinkView :
...
-            , Palette.color tonePrimary brightnessMiddle
-                |> Palette.toElementColor
-                |> Font.color
-            , Font.regular
-            , Font.underline
-            , Element.pointer
-            ]
-                ++ (ARIA.toElementAttributes <| ARIA.roleButton)
+            buttonWidth width
+                :: (Palette.color tonePrimary brightnessMiddle
```
PedroHLC, line 510: `O(n=6)` vs `6*O(1)` :joy: 
Not sure why it suggests this, maybe it's for the case we do something like `((x ++ y) ++ z)` :man_shrugging:

showcase/src/Main.elm
```diff
@@ -9,8 +9,8 @@ import Html exposing (Html, div, img)
 import Html.Attributes exposing (src, style)
 import Icons
 import Layouts.Stories as Layouts
-import LoadingView as LoadingView
-import Model as Model exposing (Model)
+import LoadingView
```
PedroHLC, line 12: Great new rule! :clap:  :clap:  :clap:

src/I18n/English.elm
```diff
@@ -1,4 +1,4 @@
-module I18n.English exposing (..)
+module I18n.English exposing (root)
```
PedroHLC, line 1: These `src/I18n` files are pre-generated, we should remove them from linters...

src/UI/Internal/NList.elm
```diff
@@ -1,41 +0,0 @@
-module UI.Internal.NList exposing (..)
```
PedroHLC, line 1: I should have put this in an external package, but I agree in deleting it :smile:

src/UI/Internal/SideBar.elm
```diff
@@ -1,6 +1,20 @@
...
 
-import Element exposing (Attribute, Element, fill, fillPortion, height, padding, paddingEach, paddingXY, px, scrollbarY, shrink, spacing, width)
+import Element
+    exposing
+        ( Attribute
+        , Element
+        , fill
+        , fillPortion
+        , height
+        , padding
```
PedroHLC, line 10: Can we add a rule so that we can have `(Element.width fill)` and `(Element.paddingXY 0 12)` but not `(width Element.fill)` neither `(paddingXY 0 12)` ?? (Import exposing right elements like fill/rgb/shrink/px but never left ones like height/width/padding)

src/UI/Internal/Tables/FiltersView.elm
```diff
@@ -177,23 +177,23 @@ headerAttrs isApplied =
...
+    Primitives.roundedBorders
+        contextSize
+        :: Element.width Element.fill
+        :: paddingXY
+        :: Element.spacing 8
+        :: Font.size textSize
+        :: Font.semiBold
+        :: Element.pointer
+        :: (ARIA.toElementAttributes ARIA.roleButton
+                ++ workingTheme
```
PedroHLC, line 195: Even here, `O(n=9)` has gone to `O(9)` :kangaroo: 

@segmentationfaulter can we do it so that:
*  `(([1, 2, 3] ++ [4, 5, 6]) ++ [7, 8, 9])` becomes `([7, 8, 9] :: 1 :: 2 :: 3 :: 4 :: 5 :: 6)`. In here `O(9) !== O(6)`
*  `([1, 2, 3] ++ ([4, 5, 6] ++ [7, 8, 9]))` continues as it's? In here `O(6) === O(6)`

src/UI/Layout/SplitSelectable.elm
```diff
@@ -20,30 +19,6 @@ type alias Config object msg =
...
 
 
-type alias MobileConfig msg =
-    { action : Maybe (Action.WithIcon msg)
-    , title : ( String, Maybe String )
-    , unselectMsg : msg
-    }
-
-
-mobile : RenderConfig -> MobileConfig msg -> Config object msg -> Nav.Content msg
```
PedroHLC, line 30: :warning:  **WARNING** :warning:  **WARNING** :warning:  **WARNING** :warning:  **WARNING** :warning: 

We're using this in WMS, as well as `MobileConfig`, the thing is that we forgot to add it to `exposed-modules` in `elm.json`

PedroHLC (changes requested):

renanpvaz (approved):

src/UI/Internal/SideBar.elm
segmentationfaulter: Yeah, I think we will have to write our custom rules to support these use cases.

PedroHLC (approved):

.github/workflows/ci.yml
```diff
@@ -39,8 +39,8 @@ jobs:
      - name: Run linter
        run: npx elm-format --validate src/ showcase/
 
-     - name: Run analyse
-       run: npx elm-analyse
+     - name: Run elm-review
+       run: yarn review
```
PedroHLC, line 43: I think `yarn run review` it's better.

jouderianjr (approved):

PR #144, opened 2020-12-29
Add renderIf utility

renanpvaz:
#### :thinking: What?
Add the `renderIf` utility 

#### :man_shrugging: Why?
Because it's needed in multiple projects already

### DIFF
--- a/src/UI/Utils/Element.elm
+++ b/src/UI/Utils/Element.elm
@@ -76,10 +76,12 @@ svg altText attrs children =
         |> Element.el (ARIA.toElementAttributes <| ARIA.roleImage altText)
 
 
+{-| Utility to conditionally render an element
 
-{- Utility to conditionally render an element -}
+    Element.row [] []
+        |> renderIf model.shouldDisplay
 
-
+-}
 renderIf : Bool -> Element msg -> Element msg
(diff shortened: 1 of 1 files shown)


PedroHLC (approved):
I approve the code itself. But I must leave a note that I'm not a big fan of this function, as Elm will always "solve" the provided "parameter", thus doing unnecessary calculations, processing and transformations even when the element is not visible at all...

renanpvaz:
@PedroHLC so what do you suggest? It would be nice if elm-format allowed for inline ifs

renanpvaz:
Do you think this can be an actual performance impairment or it's just less performant than an if/else

PedroHLC:
> @PedroHLC so what do you suggest? It would be nice if elm-format allowed for inline ifs

It would be. Laziness would perform better, but verbosity would increase too.

segmentationfaulter (approved):

jouderianjr (approved):

jouderianjr:
so, in the end (after the discussion on Slack) we decided that's ok to add, but we shouldn't be using it for big UI functions right? is everyone ok with that? @renanpvaz can you add some comment about that?  and fix the docs because the CI is complaining :D

PR #149, opened 2021-01-05
Use the elm-review github action

renanpvaz:
#### :thinking: What?
Use the lint github action and remove all the elm-review and elm-analyse stuff 

#### :man_shrugging: Why?
To avoid having duplicated config


package.json
```diff
@@ -44,7 +41,6 @@
     "parcel-plugin-asset-copier": "^1.1.0"
   },
   "devDependencies": {
-    "elm-doc-preview": "^5.0.5",
-    "elm-review": "^2.3.3"
```
renanpvaz, line 48: should this be kept? are your editors using the local dependecy?

package.json
```diff
@@ -11,9 +11,7 @@
     "showcase": "parcel showcase/public/index.html",
     "test": "elm-test",
     "i18n": "i18n-json-to-elm",
-    "percy": "percy exec -- cypress run",
-    "review": "npx elm-review src showcase review",
```
renanpvaz, line 15: if this is useful we could keep a script with
```
npx elm-review --template PaackEng/elm-review-config src

Comment (PedroHLC) [reply package.json:15]: Yeah, I didn't think about this, but I prefer to be able to run it through "npm run". Otherwise we'll have to teach everybody how to install it globally and that this command should be used to preview the same operation as the CI will do...

**Review by segmentationfaulter — APPROVED** [review]

File: .github/workflows/ci.yml [hunk 3]
```diff
@@ -33,14 +33,9 @@ jobs:
...
-       run: yarn install
-
-     - name: Run linter
-       run: npx elm-format --validate src/ showcase/
-
-     - name: Run elm-review
-       run: yarn run review
+     - uses: PaackEng/elm-lint-action@main
+       with:
+         extra-path: 'showcase'
```
PedroHLC, line 38: Great job!

PedroHLC (approved):

PR #152, opened 2021-01-12
Expose summary list item

renanpvaz:
#### :thinking: What?
Add UI.SummaryListItem to the exposed modules and document it

#### :man_shrugging: Why?
Because it's used everywhere but we forgot to expose it :sweat_smile:

### DIFF
--- a/elm.json
+++ b/elm.json
@@ -19,6 +19,7 @@
         "UI.Radio",
         "UI.RenderConfig",
         "UI.Size",
+        "UI.SummaryListItem",
         "UI.Tables.Common",
         "UI.Tables.Stateful",
         "UI.Tables.Stateless",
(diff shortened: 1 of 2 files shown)


PedroHLC (approved):

jouderianjr (approved):

jouderianjr:
@renanpvaz @PedroHLC I realized that this component is maybe a bit out of the "standard" right? like, SummaryList doesn't exactly neither need to receiver appConfig or be returned as an `Element` (this way we could control who can "use" this component). I mean, of course we would not change the API, but maybe in a future V2/refactor of this we can keep that in mind.

renanpvaz:
@jouderianjr I agree, we should refactor this one, I don't see reason for example for it to exist outside `ListView`

PR #157, opened 2021-01-27
DES-22 | update sidebar

renanpvaz:
#### :thinking: What?
- Add a new story with the sidebar variants
- Update the current sidebar to match the design
- Refactor conflicting spacing and other styles in the sidebar
- Include the non-persistent variant of the sidebar
- Add option to enable the non-persistent sidebar to the navigation container    

**Note**: this one is a doozy but please review extra carefully [here](https://github.com/PaackEng/paack-ui/pull/157/files#diff-be613155231ad86ba193a845a2521baa31c13abca43a6e5a8ace02d417ae43c1) 

#### :man_shrugging: Why?
Because the non-persistent sidebar variant is going to be used in BRMS and the standard sidebar was updated so that both look the same.

#### :pushpin: Jira Issue
[DES-22](https://paacklogistics.atlassian.net/browse/DES-22)

### DIFF
--- a/showcase/src/Sidebar/Stories.elm
+++ b/showcase/src/Sidebar/Stories.elm
@@ -1,6 +1,6 @@
 module Sidebar.Stories exposing (stories, update)
 
-import Element exposing (Element, fill, height, maximum, px)
+import Element exposing (Element, fill, height, maximum, px, width)
 import Model exposing (Model)
 import Msg exposing (Msg)
 import PluginOptions exposing (defaultWithoutMenu)
@@ -65,8 +65,13 @@ view : RenderConfig -> Model -> Element Msg
 view renderConfig model =
     Element.column [ height (px 600) ]
         [ iconsSvgSprite
-        , Sidebar.desktopColumn renderConfig Element.none <| menu model
+        , Sidebar.nonPersistentSidebar renderConfig page <| menu model
         ]
+
+
+page : Element Msg
+page =
+    storyBorder <| Element.el [ width (px 800), height (px 600) ] Element.none
 
 
 menu : Model -> Menu.Menu Msg
--- a/showcase/src/Sidebar/Stories.elm
+++ b/showcase/src/Sidebar/Stories.elm
@@ -9,11 +9,14 @@ import Sidebar.Model as SidebarModel
 import Sidebar.Msg as SidebarMsg
 import Tables.Book exposing (Book, books)
 import UI.Badge as Badge
+import UI.Icon as Icon
 import UI.Internal.Menu as Menu
 import UI.Internal.NavigationContainer
 import UI.Internal.SideBar as Sidebar
 import UI.Layout.SplitSelectable as SplitSelectable
+import UI.Link as Link
 import UI.ListView as ListView exposing (ListView)
+import UI.NavigationContainer as Nav
 import UI.Palette as Palette exposing (brightnessLighter, tonePrimary)
 import UI.RenderConfig exposing (RenderConfig)
 import UI.SummaryListItem as Summary
@@ -59,11 +62,29 @@ view : RenderConfig -> Model -> Element Msg
 view renderConfig model =
     Element.column [ height (px 600) ]
         [ iconsSvgSprite
-        , Sidebar.desktopColumn renderConfig Element.none <|
-            Menu.default
-                (Msg.SidebarStoriesMsg << SidebarMsg.ToggleSidebar)
-                model.sidebarStories.expanded
+        , Sidebar.desktopColumn renderConfig Element.none <| menu model
         ]
+
+
+menu : Model -> Menu.Menu Msg
+menu model =
+    Menu.Menu
+        (Menu.Properties (Msg.SidebarStoriesMsg << SidebarMsg.ToggleSidebar)
+            model.sidebarStories.expanded
+        )
+        { pages =
+            [ { labeledIcon = Icon.notifications "Notifications"
+              , link = Link.link "/notifications"
+              , isCurrent = False
+              }
+            , { labeledIcon = Icon.toggle "Routes"
+              , link = Link.link "/routes"
+              , isCurrent = True
+              }
+            ]
+        , actions = []
+        , logo = Nothing
+        }
 
 
 code : String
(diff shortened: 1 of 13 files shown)


showcase/src/Sidebar/Stories.elm
```diff
@@ -0,0 +1,134 @@
...
+            [ { labeledIcon = Icon.logout "Logout"
+              , action = Msg.SidebarStoriesMsg <| SidebarMsg.NoOp
+              }
+            ]
+        , logo =
+            Just
+                { hint = "Paack logo"
+                , body = Element.html <| Html.img [ src "logo.png" ] []
+                }
+        }
```
renanpvaz, line 124: We should probably update this logo right? Probably not on the same PR though

src/UI/NavigationContainer.elm
```diff
@@ -173,6 +173,7 @@ type alias Container msg =
     , title : String
     , dialog : Maybe (Dialog msg)
     , hasMenu : Bool
+    , hasPersistentSidebar : Bool
```
renanpvaz, line 176: this change will break in a lot of places, is there any alternative? If not then I suggest we include a record with the defaults so that next time including a new option doesn't break everything

src/UI/Internal/SideBar.elm
```diff
@@ -52,14 +65,77 @@ desktopColumn cfg page menu =
...
+                ]
+
+            else
+                [ ( "transform", "translate(-100%)" )
+                , ( "opacity", "0" )
+                , ( "pointer-events", "none" )
+                , ( "transition", "transform .4s, opacity .2s .4s" )
+                ]
+    in
+    Element.el (height fill :: css transitionStyles)
```
renanpvaz, line 128: if you don't like this approach we could start a discussion on animations/transitions because currently it seems like we haven't been doing anything on that end. I took a look at [elm-animator](https://package.elm-lang.org/packages/mdgriffith/elm-animator/latest/) and thought that it was waaay too much code for a simple transition, so this would be a simpler approach that trusts that the virtual dom will only patch the styles when the change happens. In some cases, like if a node changes from a `<div>` to a `<section>` between renders, it would not work but since we're using elm-ui **and** it's a pure language I don't see it happening.

src/UI/Internal/SideBar.elm
```diff
@@ -52,14 +65,77 @@ desktopColumn cfg page menu =
...
+desktopNonPersistent cfg page (Menu.Menu prop opt) =
+    let
+        overlayAlpha =
+            if prop.isExpanded then
+                0.5
+
+            else
+                0
+
+        overlay =
```
PedroHLC, line 78: Can this be moved to its own function? Looks to me that it's just complicating reading...

src/UI/Internal/SideBar.elm
PedroHLC: I don't think we had any animations until this point.
`elm-animator` looks like using a nuke to kill an ant...

By now, could we at least move all "unsafe CSS" to a separated module (`UI.Internal.Utils.Element` looks like it)? Then in the future, we could decide how to make them all safe and standardized. (Probably a call for the team-leader <@jouderianjr> to make) -- Also, talk with U{X,I}Team about what else should be animated...

src/UI/NavigationContainer.elm
PedroHLC: the whole `UI.NavigationContainer` should be rewritten, having `Sidebar` and `Dialog` as separated modules is a good beginning.

This record, for example, could fully become a with-pattern composable type...

Could you create a DES card for it?

src/UI/Utils/Element.elm
```diff
@@ -76,6 +77,11 @@ svg altText attrs children =
         |> Element.el (ARIA.toElementAttributes <| ARIA.roleImage altText)
 
 
+css : List ( String, String ) -> List (Attribute msg)
```
PedroHLC, line 80: What about moving this to `UI.Internal.Utils.Element`?

PedroHLC (approved):
A few comments, but overall looks good to me.

renanpvaz:
In case you haven't, you can see the sidebar here https://paackeng.github.io/paack-ui/#Complex%20components/Sidebar/Non-persistent

kress95 (approved):

PedroHLC (changes requested):
Just noticed now, but it's missing the Percy story and a dedicated tab for it...

src/UI/NavigationContainer.elm
renanpvaz: Done, since I think you're more acquainted with the module I'll leave it for you to add a description. This might be something worth sketching an RFC before jumping directly into a refactor.

PedroHLC (approved):
There is the pending stuff with animation, but overall LGTM

PedroHLC (approved):
Fully approved on my side.

src/UI/Internal/SideBar.elm
```diff
@@ -123,36 +182,49 @@ viewHead cfg (Menu.Menu prop _) label maybeStack =
...
+            paddingXY 16 24
+
+          else
+            paddingXY 6 22
+        , Background.color <|
+            if config.persistent then
+                Colors.gray.lightest
+
+            else
+                Colors.white
```
jouderianjr, line 222: Checking figma, I think the menu has the same color when it is expanded, figma screenshot:

Screen Shot 2021-02-01 at 11 09 18

src/UI/Utils/Element.elm
```diff
@@ -125,6 +126,59 @@ colorTransition time =
         |> List.map tuplesToStyles
 
 
+type alias Transition msg =
```
jouderianjr, line 129: I would make create a specific module for animation as it is a closed "realm"

jouderianjr (approved):
I think the Icons are not in the correct sizing, but let's check this in another pr

src/UI/Utils/Element.elm
renanpvaz: I agree but I think that it can be laid out better once it's included officially in the design system, which Sufiyan already said to be one of the goals after standardizing

showcase/src/Sidebar/Stories.elm
PedroHLC: Solved in https://github.com/PaackEng/paack-ui/pull/157

PR #165, opened 2021-02-01
Bump and update paack-ui-assets

renanpvaz:
#### :thinking: What?
Update paack-ui-assets
Bump major

#### :man_shrugging: Why?
Because a new version was published tweaking some icons

### DIFF
--- a/elm.json
+++ b/elm.json
@@ -3,7 +3,7 @@
     "name": "PaackEng/paack-ui",
     "summary": "Paack's Design System applied over Elm UI",
     "license": "BSD-3-Clause",
-    "version": "2.1.0",
+    "version": "3.0.0",
     "exposed-modules": [
         "UI.Alert",
         "UI.Badge",
--- a/elm.json
+++ b/elm.json
@@ -3,7 +3,7 @@
     "name": "PaackEng/paack-ui",
     "summary": "Paack's Design System applied over Elm UI",
     "license": "BSD-3-Clause",
-    "version": "2.1.0",
+    "version": "2.2.0",
     "exposed-modules": [
         "UI.Alert",
         "UI.Badge",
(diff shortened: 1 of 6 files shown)


PedroHLC (approved):

segmentationfaulter (approved):

PedroHLC (approved):

kress95 (approved):

elm.json
```diff
@@ -3,7 +3,7 @@
     "name": "PaackEng/paack-ui",
     "summary": "Paack's Design System applied over Elm UI",
     "license": "BSD-3-Clause",
-    "version": "2.1.0",
+    "version": "3.0.0",
```
PedroHLC, line 6: Ok, with the information you gave me in Slack I see the whole situation.
We won't need to bump to 3.0.0 if we do `NavigationContainer.withMenuStyle", supposing that all pages will use the same sidebar style...

PedroHLC (changes requested):

src/UI/NavigationContainer.elm
```diff
@@ -187,9 +188,17 @@ type Navigator page msg
 type alias NavigatorRecord page msg =
     { container : page -> Container msg
     , menu : Menu.Menu msg
+    , sidebarStyle : SidebarStyle
     }
 
 
+{-| The behavior/appearance that the sidebar will have when enabled
+-}
+type SidebarStyle
```
PedroHLC, line 197: Wouldn't expose the values through functions instead of exposing the union entries be for API purposes?
1 - The type would be opaque;
2 - Adding new values wouldn't break the final app's switch-cases;
3 - Adding new subtypes to those wouldn't break the user instancing functions.

PedroHLC (changes requested):

src/UI/NavigationContainer.elm
renanpvaz: generally I'd agree but in this case what `SidebarStyle` is really is a flag with just a teeny tiny more info, it doesnt hold any values and there's no reason to pattern match on it outside the NavigationContainer module

PedroHLC (approved):

PR #168, opened 2021-02-02
Sidebar Hotfix

renanpvaz:
#### :thinking: What?
Smol hack to force the sidebar to start closed when non-persistent
Fix the sidebar overlay that was preventing clicks on the content

#### :man_shrugging: Why?
Because these are the two issues I found testing the sidebar in a real app with the `NavigationContainer`


PedroHLC (approved):
LGTM, also, good work!

segmentationfaulter (approved):

PR #171, opened 2021-02-04
HOTFIX (again?) | Add entry point for nav state with the menu closed

renanpvaz:
#### :thinking: What?
Add an actual new entry point for the nav state instead of hacking my way around it

#### :man_shrugging: Why?
Because my past "fix" forces the menu to never open :facepalm:  

### :fire: Extra
Sorry, this time it'll work

### DIFF
--- a/src/UI/NavigationContainer.elm
+++ b/src/UI/NavigationContainer.elm
@@ -1,5 +1,5 @@
 module UI.NavigationContainer exposing
-    ( Msg, State, stateInit, stateUpdate
+    ( Msg, State, stateInit, stateInitWithClosedMenu, stateUpdate
     , Navigator, navigator
     , Container, containerMap
     , Content, contentSingle, StackChild, contentStackChild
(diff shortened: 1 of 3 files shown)


miguelcoba (approved):

kress95 (approved):

src/UI/NavigationContainer.elm
```diff
@@ -49,7 +49,7 @@ Example of usage:
 
 # Model & Update
 
-@docs Msg, State, stateInit, stateUpdate
+@docs Msg, State, stateInit, stateInitWithClosedMenu, stateUpdate
```
PedroHLC, line 52: :thinking: why not a proper `stateWithMenuOpen : Boolean -> State -> State`?

renanpvaz: because so far there's no need for a state that is forced open

PedroHLC: then, what about `stateInit |> stateWithMenuClosed` ?

renanpvaz: done

PedroHLC (approved):

PR #179, opened 2021-02-25
WMS-798 | Add Switch component

PedroHLC:
#### :thinking: What?

* Adds the new Swith component;
* Add proper ARIA role for the switch;
* (Unrelated) Fix Checkboxes' docs.

#### :man_shrugging: Why?

* Will be used in the new "Problem Solving" flow

#### :pushpin: Jira Issue

[WMS-798](https://paacklogistics.atlassian.net/browse/WMS-798)

#### :no_good: Blocked by

Not blocked.

#### :clipboard: Pending

Nothing pending.

### DIFF
--- a/showcase/src/Checkboxes/Stories.elm
+++ b/showcase/src/Checkboxes/Stories.elm
@@ -31,9 +31,6 @@ update msg model =
         Checkboxes.Box3Uncheck ->
             ( { model | box3 = False }, Cmd.none )
 
-        Checkboxes.SwitchSet newValue ->
-            ( { model | switch = newValue }, Cmd.none )
-
 
 stories : RenderConfig -> ExplorerUI
 stories renderConfig =
@@ -79,29 +76,12 @@ boxView renderConfig ({ checkboxesStories } as model) =
             (implicitMessages >> Msg.CheckboxesStoriesMsg)
             checkboxesStories.box3
             |> Checkbox.renderElement renderConfig
-        , switchView renderConfig model
         , ("Total: "
             ++ (String.fromFloat <| totalPrice checkboxesStories)
             ++ " USD"
           )
             |> Text.body1
             |> Text.renderElement renderConfig
-        ]
-
-
-switchView : RenderConfig -> Model -> Element Msg
-switchView renderConfig { checkboxesStories } =
-    let
-        label =
-            "Donate 10¢ to charity"
-    in
-    Element.row [ Element.spacing 8 ]
-        [ Switch.switch label
-            (Checkboxes.SwitchSet >> Msg.CheckboxesStoriesMsg)
-            checkboxesStories.switch
-            |> Switch.withActivatedColor Switch.success
-            |> Switch.renderElement renderConfig
-        , Element.text label
         ]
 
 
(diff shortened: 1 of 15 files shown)


showcase/src/Checkboxes/Stories.elm
```diff
@@ -30,6 +31,9 @@ update msg model =
         Checkboxes.Box3Uncheck ->
             ( { model | box3 = False }, Cmd.none )
 
+        Checkboxes.SwitchSet newValue ->
```
jouderianjr, line 34: If we are treating this as something different than a checkbox (UI.Switch), I would create another category in the showcase as well

src/UI/Switch.elm
```diff
@@ -0,0 +1,197 @@
...
+
+# Building
+
+@docs Switch, switch
+
+
+# Coloring
+
+@docs withActivatedColor
+@docs ActivatedColor, success, danger
```
jouderianjr, line 25: We are missing one state, I think we have 3 in figma

src/UI/Switch.elm
```diff
@@ -0,0 +1,197 @@
+module UI.Switch exposing
+    ( Switch, switch
```
jouderianjr, line 2: What do you think about instead of having the color as a modifier, we use as an entry point?

so instead of `Switch.switch`, we can have:

- `Switch.default`
- `Switch.success`
- `Switch.danger`

src/UI/Switch.elm
```diff
@@ -0,0 +1,197 @@
...
+                    :: boxAttrs
+                    ++ Element.transition False paddingTransition
+                )
+
+
+paddingTransition : Element.Transition msg
+paddingTransition =
+    Element.easyPadding
+        { top = 1, bottom = 1, left = 1, right = 20 }
+        { top = 1, bottom = 1, left = 20, right = 1 }
```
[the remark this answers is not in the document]
jouderianjr: I would add a transition to the color too, it would be a nice touch ⭐

kress95 (approved):

src/UI/Switch.elm
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+
+{-| Colors can be used to enforce a different kind of attention to the switch when activated.
+
+The default color is half-opaque black.
+
+-}
+type Color
+    = HalfOpaque
```
jouderianjr, line 60: I would add the prefix here for consistency

showcase/src/Switches/Stories.elm
```diff
@@ -0,0 +1,110 @@
...
+
+switchCode : String
+switchCode =
+    prettifyElmCode """
+    Element.column [ Element.spacing 8 ]
+        [ Switch.default "Toggle friendly mode"
+            Msg.Switch1Set
+            model.switch1
+            |> Switch.renderElement renderConfig
+            |> appendSwitchLabel "Friendly mode"
```
jouderianjr, line 98: I think having this `appendSwitchLabel` here confusing, it makes me think that this function comes from the Switch module.

maybe you could use the more explicit example: 

```elm
Element.row [ Element.spacing 8 ]
        [ Switch.default "Toggle friendly mode" Msg.Switch1Set model.switch1
            |> Switch.renderElement renderConfig
        , Element.text label
        ]

jouderianjr (approved):

PR #180, opened 2021-02-25
WMS-794 | Table Analytics

kress95:
#### :thinking: What?

1. Add some `Analytics` events for table filters and sorters
2.  `Effect` is now a `List (SideEffect msg)` like on `wms-web`

#### :man_shrugging: Why?

1. To provide more analytics events to `wms-web`
2. Because the previous change required firing `Analytics` along with a `MsgToCmd` effect

#### :pushpin: Jira Issue

[WMS-794](https://paacklogistics.atlassian.net/browse/WMS-794)

### DIFF
--- a/src/UI/Analytics.elm
+++ b/src/UI/Analytics.elm
@@ -14,44 +14,45 @@ type TableAnalytics
     | ClearSorting
 
 
-encode : Analytics -> Value
+type alias Property =
+    ( String, Value )
+
+
+encode : Analytics -> List Property
 encode analytics =
     case analytics of
         TableAnalytics a ->
             encodeTable a
 
 
-encodeAction : String -> ( String, Value )
+encodeAction : String -> Property
 encodeAction name =
     ( "action", Encode.string name )
 
 
-encodeColumn : Int -> ( String, Value )
+encodeColumn : Int -> Property
 encodeColumn column =
     ( "column", Encode.int column )
 
 
-encodeTable : TableAnalytics -> Value
+encodeTable : TableAnalytics -> List Property
 encodeTable analytics =
     case analytics of
         ApplyFilter column ->
-            Encode.object
-                [ encodeAction "apply_filter"
-                , encodeColumn column
-                ]
+            [ encodeAction "apply_filter"
+            , encodeColumn column
+            ]
 
         ClearFilter column ->
-            Encode.object
-                [ encodeAction "clear_filter"
-                , encodeColumn column
-                ]
+            [ encodeAction "clear_filter"
+            , encodeColumn column
+            ]
 
         SetSorting column reverse ->
-            Encode.object
-                [ encodeAction "set_sorting"
-                , encodeColumn column
-                , ( "reverse", Encode.bool reverse )
-                ]
+            [ encodeAction "set_sorting"
+            , encodeColumn column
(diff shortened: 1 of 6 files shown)


src/UI/Effect.elm
```diff
@@ -0,0 +1,58 @@
+module UI.Effect exposing
+    ( Effect
+    , SideEffect(..)
```
renanpvaz, line 3: not sure if this is the best time, but the need for a common package for effects is here. The question for me is, should this package have the definition of the `SideEffect` or should it be generic?

PedroHLC (approved):
Which approach do you recommend for us to inject which table caused this event in WMS?

src/UI/Effect.elm
kress95: - This PR is using the `Effect` type as an `OutMsg`.

It doesn't feel wrong but has some duplicated code as a result.
The alternatives are:

- Something like `Translator` the pattern (generic effects)
- Centralizing every `Effect` in a package, kinda following the `NoMap` pattern

The `Mixpanel` integration in `wms-web` is currently using the `Translator` pattern, but the API there is way simpler.

segmentationfaulter (approved):

jouderianjr (commented):
@kress95 sorry to only say this right now, but I was wondering if maybe makes sense to have the analytics through another value in the tuple instead make the table trigger `Effects`.

so the update function would be: 

```elm
update 
    : Msg item 
    -> State msg item column 
    -> (State msg item columns, Cmd msg, AnalyticsEvent)
```

this way we keep Analytics, we remove this complexity in the package. does that make sense?

renanpvaz (approved):

src/UI/Analytics.elm
```diff
@@ -0,0 +1,58 @@
+module UI.Analytics exposing (Analytics(..), TableAnalytics(..), encode)
```
jouderianjr, line 1: Do we want to expose the type definition? I think this generates breaking changes if we add another branch in the type. so maybe it's better to move the TableAnalytics type to an internal type as far I can see, the external app does not need to know the type definition as the encode function is the one that transforms to a serializable data

kress95: > Do we want to expose the type definition?

I really planned these types as public API because we may change how we encode them depending on the app or service.

This module is not supposed to know the integration (Mixpanel) or how we send data to it, so the encoder provided here is more like a suggestion than a rule.

kress95:
> @kress95 sorry to only say this right now, but I was wondering if maybe makes sense to have the analytics through another value in the tuple instead make the table trigger `Effects`.
> 
> so the update function would be:
> 
> ```elm
> update 
>     : Msg item 
>     -> State msg item column 
>     -> (State msg item columns, Cmd msg, AnalyticsEvent)
> ```
> 
> this way we keep Analytics, we remove this complexity in the package. does that make sense?

It would make things simpler for now, but later we would have to implement effects anyway to test flows inside paack-ui.

The way wms is using this right now is directly converting the paack-ui effect to the local effect type, so even the analytics are converted to effects right away.

Also, about mixing analytics and effects in the same type, it's not that analytics are effects, it is that both are external messages that require an interpreter (a perform-ish function) to become useful (e.g. a Cmd).

PR #189, opened 2021-03-10
WMS-800 | Update ListView style

kress95:
#### :thinking: What?

Update ListView to be more consistent with [figma](https://www.figma.com/file/DTyoSveNksHpZCdU8a7BNs/Dashboard?node-id=104%3A1804):

image

#### :mag: How it looks?

image

#### :man_shrugging: Why?

It's part of the problem-solving flow.

#### :pushpin: Jira Issue

[WMS-800](https://paacklogistics.atlassian.net/browse/WMS-800)

### :fire: Extra

- I'll Include the input icon in the next PR.

### DIFF
--- a/i18n/English.json
+++ b/i18n/English.json
@@ -53,7 +53,8 @@
     "sidebar": {
         "expand": "Expand sidebar",
         "collapse": "Minimize sidebar",
-        "previous": "Go back"
+        "previous": "Go back",
+        "moreActions": "More actions"
     },
     "contentPlaceholders": {
         "nothingToSeeHere": {
--- a/i18n/French.json
+++ b/i18n/French.json
@@ -53,7 +53,8 @@
     "sidebar": {
         "expand": "Elargir la barre latérale",
         "collapse": "Réduire la barre latérale",
-        "previous": "Retourner"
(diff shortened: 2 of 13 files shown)


renanpvaz:
Few things first: 
- the card could've been created in the design system board
- the figma link in these cases could've been available in the description or the card at least
- a picture of what it looks like now would be nice

renanpvaz (approved):
oops, I understood from the title that this was going to be a full update including the new features for the list view seen in figma

PedroHLC:
Also, could you add the headers to the showcase? There isn't a ListView page, but the Layout shows one: http://localhost:1234/#Basics/Layouts/SplitSelectable

PedroHLC (approved):

package.json
```diff
@@ -44,5 +44,8 @@
   "devDependencies": {
     "elm-doc-preview": "^5.0.5",
     "elm-review": "^2.4.4"
+  },
+  "resolutions": {
```
PedroHLC, line 48: We won't need this if we merge #192  first.
But it's also an option to merge this one first and add a commit to #192 removing this.

PR #195, opened 2021-03-23
WMS-802 | Add checkbox support to SummaryListItem

kress95:
#### :thinking: What?

- Add `UI.ListView.SummaryItem`:
  - roughly the same as `UI.SummaryListItem` but using "with pattern"
  - optional badge using `withBadge`
  - optional checkbox using `withCheckbox`
  - consistent with [Figma](https://www.figma.com/file/DTyoSveNksHpZCdU8a7BNs/Dashboard?node-id=1813%3A96)
- Add `UI.ListView.simpleList`:
  - basically the same as `UI.ListView.selectList` but without built-in selection
  - no visual changes
- Update `Layouts.Stories` to use `UI.ListView.SummaryItem`

#### 👀 Results

- When not using checkbox:

image

- When using checkbox and removing selection's background:

image

#### :man_shrugging: Why?

- [Figma](https://www.figma.com/file/DTyoSveNksHpZCdU8a7BNs/Dashboard?node-id=1813%3A74)
  - changes `UI.SummaryListItem` paddings
  - adds checkboxes to `UI.SummaryListItem`, and that would introduce breaking changes without using "with pattern"

#### :pushpin: Jira Issue

[WMS-802](https://paacklogistics.atlassian.net/browse/WMS-802)

### DIFF
--- a/src/UI/V2/SummaryListItem.elm
+++ b/src/UI/V2/SummaryListItem.elm
@@ -7,7 +7,7 @@ module UI.V2.SummaryListItem exposing
 {-| `SummaryListItem` represents a single item in a list that is usually selectable or searchable.
 This component is meant to be used with [`ListView`](UI-ListView).
 
-    import UI.SummaryListItem as Summary
+    import UI.V2.SummaryListItem as Summary
 
     listItemView : AppConfig -> Bool -> Item -> Element Msg
     listItemView appConfig isSelected item =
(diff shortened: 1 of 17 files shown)


PedroHLC (commented):
Pff, I reviewed this PR yesterday, but forgot to publish the review, here it's:

src/UI/V2/SummaryListItem.elm
```diff
@@ -0,0 +1,244 @@
+module UI.V2.SummaryListItem exposing
+    ( SummaryListItem, summaryListItem
+    , withSelected, withBadge, withCheckbox
+    , renderElement
+    )
+
```
renanpvaz, line 6: what do you think about creating this as `UI.ListView.SummaryItem` or something on that line? So that later we don't have to change the modules again when we get rid of the original summary item and also this way it makes sense considering that both components are tied.

renanpvaz (approved):
great work!! :1st_place_medal:

PedroHLC (approved):

PR #204, opened 2021-04-13
Add V2.Layout.SplitSelectable

kress95:
#### :thinking: What?

Adds a V2.Layout.SplitSelectable with included support to put elements before the layout.

#### :man_shrugging: Why?

Required for [Fix Issues status bar](https://www.figma.com/file/DTyoSveNksHpZCdU8a7BNs/Dashboard?node-id=1813%3A8156).

### :fire: Extra

I'm not entirely happy with creating this V2, but adding with pattern also didn't feel right, maybe this element should be internal to `wms-web`.

### DIFF
--- a/elm.json
+++ b/elm.json
@@ -37,6 +37,7 @@
         "UI.Utils.Focus",
         "UI.Utils.TypeNumbers",
         "UI.V2.Dialog",
+        "UI.V2.Layout.SplitSelectable",
         "UI.Layout.SplitSelectable"
     ],
     "elm-version": "0.19.0 <= v < 0.20.0",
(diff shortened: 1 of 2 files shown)


renanpvaz:
This element was initially part of wms web. I have a feeling that perhaps layouts should be a set of elements that can be assembled instead of having a single entry point that renders the whole thing

kress95:
> This element was initially part of wms web. I have a feeling that perhaps layouts should be a set of elements that can be assembled instead of having a single entry point that renders the whole thing
> 
> The layout should return plain elements instead of `Nav.Content`, that way the status bar could be added in wms without having to touch the layout, because it isn't something internal either way

I think this is the correct path to follow, but we should better think/discuss how to do it properly.

For now, I'll close this and move this file to `Pages.FixIssues.Layout`.

PR #205, opened 2021-04-13
WMS-803 | ListView's withCustomExtraMenu

kress95:
#### :thinking: What?

Adds support to include a custom dropdown that is toggled by `moreActions` (ellipsis) button.

#### :man_shrugging: Why?

Required by Fix Issues:

image

#### :pushpin: Jira Issue

[WMS-803](https://paacklogistics.atlassian.net/browse/WMS-803)

### DIFF
--- a/src/UI/ListView.elm
+++ b/src/UI/ListView.elm
@@ -312,6 +312,7 @@ withActionBar config (SelectList prop opt) =
 
     ListView.withCustomDropdown customDropdownView
         someListView
+
 -}
 withCustomDropdown : msg -> Bool -> Element msg -> ListView object msg -> ListView object msg
 withCustomDropdown toggleMsg isEnabled body (SelectList prop opt) =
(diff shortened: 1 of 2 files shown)


src/UI/ListView.elm
```diff
@@ -199,6 +200,14 @@ type alias ToggleableCover =
     ToggleableList.Cover
 
 
+type alias Dropdown msg =
+    { toggleMsg : msg, isEnabled : Bool, body : DropdownBody msg }
```
PedroHLC, line 204: What do you think of making it opaque and with a construction function, so in the future, we can use "with-pattern" with it?

src/UI/ListView.elm
```diff
@@ -299,6 +308,20 @@ withActionBar config (SelectList prop opt) =
     SelectList prop { opt | actionBar = Just config }
 
 
+{-| Adds a button to toggle a custom dropdown element.
+
+    ListView.withCustomDropdown customDropdownView
+        someListView
+
+-}
+withCustomDropdown : msg -> Bool -> Element msg -> ListView object msg -> ListView object msg
```
PedroHLC, line 317: In theory that's a Context Menu, and in this case, one hidden in an "extra button". What do you think of calling "withExtraMenu"?

kress95: Maybe `withCustomExtraMenu`.

Because we may better define this layout later and add an `withExtraMenu : DropdownMenu msg -> ListView object msg -> ListView object msg`.

src/UI/ListView.elm
kress95: It is opaque, this type is not exposed at all.

renanpvaz (approved):

PedroHLC (approved):

PR #225, opened 2021-05-25
WMS-916 | SummaryItem tone, icon and bullet icon

kress95:
#### :thinking: What?

Adds the following features to `SummaryItem`:

- Entry tone using `withTone`
- Suffix icon using `withIcon`
- Bullet with icon using `withBulletIcon`

#### :man_shrugging: Why?

Required for problem-solving:

- [The red wrench icon:](https://www.figma.com/file/DTyoSveNksHpZCdU8a7BNs/Dashboard?node-id=2330%3A19005)

image

- [The background tone and yellow bullet with icon:](https://www.figma.com/file/DTyoSveNksHpZCdU8a7BNs/Dashboard?node-id=1813%3A191)

image

#### 🔍 How it looks?

image

image

#### :pushpin: Jira Issue

[WMS-916](https://paacklogistics.atlassian.net/browse/WMS-916)

#### :no_good: Blocked by

- [ ] https://github.com/PaackEng/paack-ui/pull/226
 
### :fire: Extra

Spent some extra time trying to properly align that suffix icon, but couldn't without resorting to some ugly hacks including hardcoded sizes. I really think we should ignore that for now and try to align it later...

### DIFF
--- a/src/UI/ListView/SummaryItem.elm
+++ b/src/UI/ListView/SummaryItem.elm
@@ -1,6 +1,6 @@
 module UI.ListView.SummaryItem exposing
     ( SummaryItem, summaryItem
-    , withSelected, withBadge, withCheckbox
+    , withSelected, withBadge, withCheckbox, withIcon, withBulletIcon, withBackgroundTone
     , renderElement
     )
 
@@ -32,7 +32,7 @@ This component is meant to be used with [`ListView`](UI-ListView).
 
 # Options
 
-@docs withSelected, withBadge, withCheckbox
+@docs withSelected, withBadge, withCheckbox, withIcon, withBulletIcon, withBackgroundTone
 
 
 # Rendering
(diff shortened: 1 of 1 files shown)


src/UI/ListView/SummaryItem.elm
```diff
@@ -103,6 +112,39 @@ withSelected selected (SummaryItem prop opt) =
...
+
+{-| Applies a tone to summary's background.
+
+    Summary.withTone Palette.toneWarning
+        someListView
+
+-}
+withTone : Palette.Tone -> SummaryItem msg -> SummaryItem msg
+withTone tone (SummaryItem prop opt) =
+    SummaryItem prop { opt | tone = Just tone }
```
renanpvaz, line 145: why not just take a `Palette.Color`? Also I think "tone" is not a common term for this kind of stuff and I can't tell from just looking at it if it changes the font color or the background, so I suggest going with something like `withBackground` or `withBackgroundColor` which says exactly what it is

src/UI/ListView/SummaryItem.elm
```diff
@@ -103,6 +112,39 @@ withSelected selected (SummaryItem prop opt) =
...
 
+{-| Adds an icon to the right side of the summary.
+
+    Summary.withIcon (Icon.fix "Problem Solving")
+        someListView
+
+-}
+withIcon : Icon -> SummaryItem msg -> SummaryItem msg
+withIcon icon (SummaryItem prop opt) =
+    SummaryItem prop { opt | icon = Just icon }
```
renanpvaz, line 123: if we see ourselves changing the API in the future to receive yet another thing to be put to the right, perhaps we should start calling it an "addon" and just receive any element

src/UI/ListView/SummaryItem.elm
kress95: > why not just take a `Palette.Color`

Mainly because the tone is reused internally with different brightnesses

PedroHLC (approved):

renanpvaz (approved):

PR #226, opened 2021-05-25
Update colour palettes

kress95:
#### :thinking: What?

Update colour palettes.

Breaks the way we refer to light and dark shades internally because we have "lighter than lightest" shades being used around.

#### :man_shrugging: Why?

[Problem-solving requires the fourth light shade of yellow](https://www.figma.com/file/T885YF0d8G5yWuI9EC49rc/Colors?node-id=0%3A1)

### :fire: Extra

Maybe revert the `darkest` changes.

### DIFF
--- a/showcase/src/Sizes.elm
+++ b/showcase/src/Sizes.elm
@@ -72,8 +72,8 @@ icons =
 iconView : RenderConfig -> Size -> ( String -> Icon, String ) -> Element msg
 iconView cfg size ( iconFn, label ) =
     Element.column
-        [ Background.color Colors.gray.darkest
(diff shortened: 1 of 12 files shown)


PedroHLC:
> Colour? Are ye British?
:sweat_smile: I think he copied my Grammarly settings. It corrects to "Colour" and "Behaviour" instead of "Color" and "Behavior"...

kress95:
> > Colour? Are ye British?
> 
> 😅 I think he copied my Grammarly settings. It corrects to "Colour" and "Behaviour" instead of "Color" and "Behavior"...

Actually, I have my own setting :joy:

PedroHLC (approved):

PR #238, opened 2021-06-07
Fix header

kress95:
#### :thinking: What?

Fix header width bug that is making the header render one word per line.

#### :man_shrugging: Why?

The header badge feature introduced along with `3.5.0` changed `Element.width fill` to `Element.width shrink`, this is causing the header to render one word per line. 

image

This temporary fix should be enough to maintain current styling in pages that aren't using header badges.


PedroHLC (approved):
LGTM

renanpvaz (approved):

PR #244, opened 2021-06-14
DES-100 | Dropdown

kress95:
#### :thinking: What?

Adds dropdown component.

#### 🔎 How it looks?

- Basic dropdown

image

- Filterable dropdown

image

#### :man_shrugging: Why?

[Figma:](https://www.figma.com/file/DTyoSveNksHpZCdU8a7BNs/Dashboard?node-id=2666%3A44774)

image

#### :pushpin: Jira Issue

[DES-100](https://paacklogistics.atlassian.net/browse/DES-100)

#### :clipboard: Pending

Some colors aren't 1:1, but they aren't present in the palette.

Maybe I should hardcode them for now 🤔 

Also, I could not get the font weights right for a screenshot

### :fire: Extra

I spent quite some time on this, but still, I'm not happy with the outcome.

The main problem is with the update requiring `RenderConfig` and `Dropdown item msg`, we should refactor `elm-ui-dropdown` to fix that.

Another thing, the major bump to `4.0.0` is there because there's a breaking change in `UI.Effect` (a new `DomFocus` effect).

### DIFF
--- a/src/UI/Dropdown.elm
+++ b/src/UI/Dropdown.elm
@@ -44,7 +44,6 @@ import Dropdown
 import Element exposing (Attribute, Element)
 import Element.Background as Background
 import Element.Border as Border
-import Element.Font as Font
 import UI.Effect as Effect exposing (Effect)
 import UI.Icon as Icon
 import UI.Internal.Colors as Colors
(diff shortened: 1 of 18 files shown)


PedroHLC (changes requested):
Missing new story page in Percy

PedroHLC (approved):
>  Maybe I should hardcode them for now thinking

I disagree, once the UI guys are done with the new color palette, we review them here.

src/UI/Dropdown.elm
```diff
@@ -0,0 +1,437 @@
...
+    Dropdown.withPlaceholder "Choose a book"
+        someDropdown
+
+-}
+withFilterPlaceholder : String -> Dropdown item msg -> Dropdown item msg
+withFilterPlaceholder placeholder (Dropdown prop opt) =
+    Dropdown prop { opt | filterPlaceholder = Just placeholder }
+
+
+{-| Replaces the component's list of elements.
```
renanpvaz, line 236: I don't think we should use the word "replaces" unless there's a relevant default value to be considered and even in that case we could use "override", otherwise I suggest writing "change" or "set"

kress95: > Looks really good, my only other complaint is that the font size used for the input in the filterable version looks too big compared to the rest ("Choose a book" in the image)
> 
> image
> 
> I couldn't find an example of the filterable dropdown having a different font size in figma, so I'm assuming it's the same

We'll have to [change `elm-ui-dropdown` to fix that](https://github.com/PaackEng/elm-ui-dropdown/blob/master/src/Dropdown.elm#L680-L681), I think `withFilterPlaceholder` should receive a [`Input.Placeholder`](https://package.elm-lang.org/packages/mdgriffith/elm-ui/latest/Element-Input#Placeholder) instead of a `String`.

kress95:
> Looks really good, my only other complaint is that the font size used for the input in the filterable version looks too big compared to the rest ("Choose a book" in the image)
> 
> image
> 
> I couldn't find an example of the filterable dropdown having a different font size in figma, so I'm assuming it's the same

We'll have to [change `elm-ui-dropdown` to fix that](https://github.com/PaackEng/elm-ui-dropdown/blob/master/src/Dropdown.elm#L680-L681), I think `withFilterPlaceholder` should receive a [`Input.Placeholder`](https://package.elm-lang.org/packages/mdgriffith/elm-ui/latest/Element-Input#Placeholder) instead of a `String`.

PedroHLC:
>  I think `withFilterPlaceholder` should receive a [`Input.Placeholder`](https://package.elm-lang.org/packages/mdgriffith/elm-ui/latest/Element-Input#Placeholder) instead of a `String`.

Nah, don't break it, make a "withPlaceholderAttributes". That's why we have with-pattern to begin with :sweat_smile:

kress95:
@renanpvaz

>  my only other complaint is that the font size used for the input in the filterable version looks too big

I'm going to fix that once `withPlaceholderAttributes` is available.

PR #246, opened 2021-06-15
DES-108 | Update checkbox component

renanpvaz:
#### :thinking: What?

- Update the Checkbox component to match the design
- Fix accessibility issues with the Checkbox
- Create a `UI.Internal.SelectionControl` to share what's common between radios and checkboxes
- Update the List View "select all" button to be a checkbox
 
#### :man_shrugging: Why?

https://www.figma.com/file/rqAoj2WwQZYoisP2G6d6nJ/Selection-Controls

#### :pushpin: Jira Issue

[DES-108](https://paacklogistics.atlassian.net/browse/DES-108)

#### :no_good: Blocked by

Nothing.

#### :clipboard: Pending

- [x] Update the label visible behavior
- [x] Update other internal places that use the checkbox (e.g. ListView's select all) 
- [x] Fix wrapping labels
- [x] (Partial) Fix bug where clicking the label selects all radio buttons
- [ ] Determine whether the selection control width should be customizable

### :fire: Extra

Both the radio and the checkboxes had no support for keyboard navigation and (I guess) poor accessibility overall. I noticed that the switch component has issues too so maybe we should look into auditing our components for accessibility issues, thoughts?


src/UI/Checkbox.elm
```diff
@@ -70,65 +97,90 @@ type alias Options =
...
 
-    TextField.setLabelVisible True someTextField
+    Checkbox.withHiddenLabel someCheckbox
 
 -}
-withLabelVisible : Bool -> Checkbox msg -> Checkbox msg
-withLabelVisible bool (Checkbox prop opt) =
-    Checkbox prop { opt | labelVisible = bool }
+withHiddenLabel : Checkbox msg -> Checkbox msg
+withHiddenLabel (Checkbox prop opt) =
```
renanpvaz, line 109: I noticed that the sole purpose of this function was to hide the label when we only need the checkbox input appearing so I changed it to constrain it to that specific case. If you know a place where we need to toggle the label to be visible for some reason or you disagree to the change let me know.

src/UI/ListView.elm
```diff
@@ -596,85 +596,13 @@ selectAllButtonView : RenderConfig -> Maybe (SelectAll msg) -> Element msg
...
-                    |> Text.withColor color
-                    |> Text.renderElement cfg
-                , checkIcon
-                ]
-                |> Element.el
-                    [ Element.paddingEach
-                        { top = 0, bottom = 12, left = 12, right = 12 }
-                    ]
+            Checkbox.checkbox (cfg |> localeTerms >> .listView >> .selectAll) message state
+                |> Checkbox.renderElement cfg
```
renanpvaz, line 600: Figma was updated to use the checkbox according to the design system: https://www.figma.com/file/DTyoSveNksHpZCdU8a7BNs/Dashboard?node-id=2449%3A27203

src/UI/Internal/SelectionControl.elm
```diff
@@ -0,0 +1,69 @@
...
+            Sizes 20 8 2
+
+        SizeMD ->
+            Sizes 28 10 3
+
+
+buttonAttributes : SelectionControlSize -> List (Element.Attribute msg)
+buttonAttributes size =
+    [ Element.spacing 10
+    , Element.width shrink
```
renanpvaz, line 41: Please check what this looks like for both checkboxes and radio (better seen locally focusing on them) and let me know  whether you think the width should be configurable or not

miguelcoba (approved):

kress95 (approved):

elm.json
```diff
@@ -3,7 +3,7 @@
     "name": "PaackEng/paack-ui",
     "summary": "Paack's Design System applied over Elm UI",
     "license": "BSD-3-Clause",
-    "version": "4.0.0",
+    "version": "5.0.0",
```
PedroHLC, line 6: Oshino Meme

PedroHLC, line 6: Oshino Meme

renanpvaz: image

PedroHLC: Hey @renanpvaz , now that we're on this subject, `RadioSize` should looked like:
- https://github.com/PaackEng/elm-google-maps/blob/ed273d827f549459297a497b0e41f512ef78b529/src/GoogleMaps/Map.elm#L91
- https://github.com/PaackEng/elm-google-maps/blob/ed273d827f549459297a497b0e41f512ef78b529/src/GoogleMaps/Map.elm#L172
- https://github.com/PaackEng/elm-google-maps/blob/ed273d827f549459297a497b0e41f512ef78b529/src/GoogleMaps/Map.elm#L188

src/UI/Internal/SelectionControl.elm
PedroHLC: I liked the "shrink", I don't see in this combination (radio/check ++ text label) a reason for it to fill.
As this is just an "update" for already-existence implementation, we can go with shrink, and then for each new occurrence ask the UI team what should be the behavior...

PedroHLC (approved):
Note that you somehow affected the table's rows' height.

PR #248, opened 2021-06-18
Fix radio button

renanpvaz:
#### :thinking: What?

- Add html ids to the radio buttons
- Update the messaging to send the id when selecting an option
- Update stories to handle the id and focus on the button 

#### :man_shrugging: Why?

Because the focus was not following the selection

#### :pushpin: Jira Issue

[DES-106](https://paacklogistics.atlassian.net/browse/DES-106)

#### :no_good: Blocked by

Nothing.

#### :clipboard: Pending

- [ ] Feed the trolls.

### :fire: Extra

Not happy at all having to break again the API and also the fact that the burden of focusing on the element lies on who's using it. I'm also not happy having to "generate" an id on the spot, although it should be reliable considering no one should have options with the same label. Any alternatives will be considered, these are some of them:
- Creating a `Radio.update : Radio.Msg -> (RadioState, Effect msg)` and returning a focus effect
- Creating a new entry point of the component as an alternative to `Radio.group` so that we at least avoid breaking the API
- A separate optional msg to handle the id when selecting (not sure how to do this though)
- Use real inputs and labels instead of the elm-ui implementation

To me this seems like [yet another](https://github.com/mdgriffith/elm-ui/issues/318) bug with elm-ui's radio: if the options are focusable it doesn't make for the focus not to follow the selection.

### DIFF
--- a/elm.json
+++ b/elm.json
@@ -3,7 +3,7 @@
     "name": "PaackEng/paack-ui",
     "summary": "Paack's Design System applied over Elm UI",
     "license": "BSD-3-Clause",
-    "version": "5.0.0",
+    "version": "6.0.0",
     "exposed-modules": [
         "UI.Alert",
         "UI.Analytics",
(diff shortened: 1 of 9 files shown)


kress95 (approved):

segmentationfaulter (approved):

PedroHLC (approved):
Just remember to correctly approach multiple radio groups on the same page.

PR #261, opened 2021-07-12
LMO-515 | DES-111 | Prepare reusable Filter component (Internals)

PedroHLC:
<div align="center">joke</div>

#### :thinking: What?

- Bump paack-ui-assets;
- Add filter component's internals;
- Move table's filter to the component's internal;
- Update style to match Figma;
- Allow sorting with more than alphabetically;
- Add notes about "Fira Sans" weights;
- Fix radio's button text-style.

#### :man_shrugging: Why?

- First part of what I need in LMO;
- Fresh look.

#### :pushpin: Jira Issue

[DES-111](https://paacklogistics.atlassian.net/browse/DES-111) and [LMO-515](https://paacklogistics.atlassian.net/browse/LMO-515).

#### :no_good: Blocked by

Nothing.

#### :clipboard: Pending

In a future PR:
- The API to the Final reusable Filter;
- headerWithFilter / bodyWithFilter;
- Fix of the date error state;
- Fix of the "portional-width" minimal size.

### DIFF
--- a/src/UI/Internal/Filter/Model.elm
+++ b/src/UI/Internal/Filter/Model.elm
@@ -1,0 +1,635 @@
+module UI.Internal.Filter.Model exposing (..)
+
+import Array exposing (Array)
+import Time exposing (Posix)
+import UI.Internal.Basics exposing (flip, maybeNotThen)
+import UI.Internal.DateInput as DateInput exposing (DateInput(..), PeriodComparison(..), PeriodDate, RangeDate)
+
+
+
+-- Constant
+
+
+dateSeparator : String
+dateSeparator =
+    "/"
+
+
+
+-- Main Type
+
+
+type Filter msg item
+    = SingleTextFilter (SingleTextFilterConfig msg item)
+    | MultiTextFilter (MultiTextFilterConfig msg item)
+    | SingleDateFilter (SingleDateFilterConfig msg item)
+    | RangeDateFilter (RangeDateFilterConfig msg item)
+    | PeriodDateFilter (PeriodDateFilterConfig msg item)
+    | SelectFilter (List String) (SelectFilterConfig msg item)
+
+
+type alias Editable value =
+    { current : Maybe value
+    , applied : Maybe value
+    }
+
+
+type alias FilterConfig msg value item =
+    { editable : Editable value
+    , strategy : Strategy msg value item
+    }
+
+
+type Strategy msg value item
+    = Local (item -> value -> Bool)
+    | Remote (RemoteMessages msg value)
+
+
+type alias RemoteMessages msg value =
+    { applyMsg : value -> msg, clearMsg : msg }
+
+
+configSetEditable : FilterConfig msg value item -> Editable value -> FilterConfig msg value item
+configSetEditable config newEditable =
+    { config | editable = newEditable }
+
+
+
+-- Editable
+
+
+editableEmpty : Editable something
+editableEmpty =
+    { current = Nothing, applied = Nothing }
+
+
+editableWithDefault : data -> Editable data -> data
+editableWithDefault default { current, applied } =
+    current
+        |> maybeNotThen applied
+        |> Maybe.withDefault default
+
+
+editableApply : Editable data -> Editable data
+editableApply { current } =
+    { current = Nothing, applied = current }
+
+
+editableSetCurrent : value -> Editable value -> Editable value
+editableSetCurrent new old =
+    { old | current = Just new }
+
+
+editableMapCurrent : (Maybe value -> value) -> Editable value -> Editable value
+editableMapCurrent applier old =
+    { old | current = Just <| applier <| maybeNotThen old.applied <| old.current }
+
+
+
+-- Strategy
+
+
+strategyLocal : (item -> value -> Bool) -> Strategy msg value item
+strategyLocal isKept =
+    Local isKept
+
+
+strategyRemote : RemoteMessages msg value -> Strategy msg value item
+strategyRemote messages =
+    Remote messages
+
+
+
+-- Get Filters
+
+
+isEdited : Filter msg item -> Bool
+isEdited filter =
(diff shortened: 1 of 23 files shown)


miguelcoba (approved):
Great job

src/UI/Internal/Filter/View.elm
```diff
@@ -0,0 +1,520 @@
...
+    let
+        width =
+            if common.width == shrink then
+                Element.tuplesToStyles ( "min-width", "100%" )
+
+            else
+                Element.width common.width
+
+        attrs =
+            [ Element.zIndex 9
```
renanpvaz, line 269: why 9? maybe it's good to keep track of these values as constants

src/UI/Internal/Filter/View.elm
```diff
@@ -0,0 +1,520 @@
...
+
+sortAs :
+    RenderConfig
+    -> SortingDirection
+    -> { fontSize : Int, iconSize : Int, padding : RectangleSides }
+    -> BodySorting msg
+    -> String
+    -> msg
+    -> Element msg
+sortAs renderConfig direction { fontSize, iconSize } sortingData label msg =
```
renanpvaz, line 382: I think this could be a single record

src/UI/Internal/Filter/View.elm
```diff
@@ -0,0 +1,520 @@
...
+                    ]
+
+                ( Just { smaller, larger }, SortDescending ) ->
+                    [ Element.text label
+                    , Element.text " ("
+                    , Element.text larger
+                    , Element.text " - "
+                    , Element.text smaller
+                    , Element.text ")"
+                    ]
```
renanpvaz, line 411: can we reuse these with a small inline function?

src/UI/Internal/Filter/View.elm
```diff
@@ -0,0 +1,520 @@
...
+                |> Icon.renderElement renderConfig
+                |> Element.el attrs
+
+        Just SortDescending ->
+            RenderConfig.localeTerms renderConfig
+                |> (.tables >> .sorting >> .descending)
+                |> Icon.sortDecreasing
+                |> Icon.withCustomSize iconSize
+                |> Icon.renderElement renderConfig
+                |> Element.el attrs
```
renanpvaz, line 512: can we reuse these as well?

src/UI/Internal/Utils/Element.elm
```diff
@@ -56,29 +58,31 @@ id value =
...
 
-overlayBackground : msg -> Element msg
-overlayBackground onClickMsg =
+clickElsewhereToLeave : msg -> List (Attribute msg) -> Element msg -> Element msg
+clickElsewhereToLeave onClickMsg backgroundStyleAttributes foregroundContent =
+    let
+        backgroundAttributes =
+            positionFixed
+                -- Needs for starting at the top-left corner
+                :: zIndex 8
```
renanpvaz, line 67: Will this zindex work for any page? Maybe it should be passed in and stored as a constant like I suggested previously. To be more specific, giving it a name like mySpecificElementsNameZIndex, not zIndex8

src/UI/Internal/Utils/Element.elm
```diff
@@ -56,29 +58,31 @@ id value =
         |> Element.htmlAttribute
 
 
-overlayBackground : msg -> Element msg
-overlayBackground onClickMsg =
+clickElsewhereToLeave : msg -> List (Attribute msg) -> Element msg -> Element msg
+clickElsewhereToLeave onClickMsg backgroundStyleAttributes foregroundContent =
```
renanpvaz, line 62: TBH I prefer the old name, this name gives me the impression that it should be of type `Attribute msg` instead of an element

renanpvaz:
I tested it locally and it looks good! I only found a few things: 
- The shadow looks too dark, I think we should use the one from figma
- There's no thin gray border around the filter popover like in the design
- Selecting a sorting closes the filter but applying or clearing them doesn't
- The text fields in the filter start in the error state

image

src/UI/Internal/Filter/Model.elm
```diff
@@ -0,0 +1,657 @@
...
+            editable.current /= Nothing
+
+        RangeDateFilter { editable } ->
+            editable.current /= Nothing
+
+        PeriodDateFilter { editable } ->
+            editable.current /= Nothing
+
+        SelectFilter _ { editable } ->
+            editable.current /= Nothing
```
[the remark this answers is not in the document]
PedroHLC: `current` in those entries variates between `Maybe String`, `Maybe (List String)`, and `Maybe Int`.
Elm does not provide me any way for me to simplify this...

src/UI/Internal/Filter/View.elm
PedroHLC: Sorry, I disagree:
1 - It's not confusing as every entry has a different type;
2 - Wrongly inserting one argument would break the compilation;
3 - A record would increase overhead;
4 - A record would break the capability for partially applying arguments;

src/UI/Internal/Utils/Element.elm
PedroHLC: What do you think of "transparentOverlay" ?

PedroHLC:
> The shadow looks too dark, I think we should use the one from figma

I accidentally made one shadow of 50% alpha, when there were actually two of 5&.
I don't think it's worth perform-wise to have two shadows, so I fixed it with a single one of 9%.

> The text fields in the filter start in the error state

Oh yeah, it starts empty, and empty are not valid dates :smile:

renanpvaz:
> Oh yeah, it starts empty, and empty are not valid dates

empty should be valid because it means no filter, right? It's weird for the user to open it a first time and see an error

renanpvaz (approved):
Just missing the initial invalid state of the text field filters, the rest looks good tho

PR #274, opened 2021-08-03
Dropdown scrollbar

kress95:
#### :thinking: What?

Limits the dropdown part of the Dropdown component to have at most `200px` of height and adds a scrollbar for longer dropdowns.

> also some minor documentation fixes

#### :man_shrugging: Why?

The current dropdown has no height limit, this can cause issues with or long lists and/or shorter displays.

### :fire: Extra

Change `PaackEng/elm-ui-dropdown` to send `DomFocus` effects when selection change.

### DIFF
--- a/src/UI/Dropdown.elm
+++ b/src/UI/Dropdown.elm
@@ -2,6 +2,7 @@ module UI.Dropdown exposing
     ( Dropdown, BasicConfig, basic, filterable
     , State, Msg, init, update
     , withPlaceholder, withFilterPlaceholder, withItems, withSelected, withItemToText
+    , withMaximumListHeight
     , renderElement
     )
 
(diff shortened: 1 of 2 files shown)


src/UI/Dropdown.elm
```diff
@@ -391,6 +391,8 @@ selectAttrs (Dropdown prop _) =
 listAttrs : List (Attribute msg)
 listAttrs =
     [ Element.width Element.fill
+    , Element.height <| px 200
```
PedroHLC, line 394: This is not limiting it, it's forcing a fixed value :eyes: 
Think we can manage to make it `shrink |> maximum XXX` ?

what about having a `withMaximumListHeight : Int -> ...`?

PedroHLC (approved):
LGTM

renanpvaz (approved):

PR #283, opened 2021-08-13
Add `Filter` support to `ListView`

kress95:
#### :thinking: What?

- Adds support for additional filters in the `ListView`
  - This change required adding `withAlignRight` to the `UI.Filter`
    - ... and I ended up fixing an unknown issue with the `UI.Filter` button width.

#### :man_shrugging: Why?

~~the voice inside my head asked it~~

Required for [WS-12](https://paacklogistics.atlassian.net/browse/WS-12)

#### 🔍 How it looks like?

Depends on usage, but it looks great like this:

image

image

```elm
-- showcase/src/Layouts/Stories.elm:127
|> ListView.withFilter
    (model.extraFilter
        |> Filter.fromModel "Filters"
            (LayoutsMsg.ForFilter >> Msg.LayoutsStoriesMsg)
        |> Filter.withWidth (Element.minimum 150 Element.fill)
    )
```

~~Though I'm not sure why the checkbox is shrinking there...~~ **FIXED**

#### :pushpin: Jira Issue

[WS-12](https://paacklogistics.atlassian.net/browse/WS-12)

### :fire: Extra

Both I and @PedroHLC are very unhappy with how we have to extend the list view for everything every time.

I wanted to improve that with a modular v2 `ListView`, but it would be a huge change and I'm worried about it requiring more boilerplate for using doing the same stuff we already do...

### DIFF
--- a/src/UI/Filter.elm
+++ b/src/UI/Filter.elm
@@ -1,6 +1,6 @@
 module UI.Filter exposing
     ( Filter
-    , FilterSize, withSize, sizeMedium, sizeExtraSmall, withWidth
+    , FilterSize, withSize, sizeMedium, sizeExtraSmall, withWidth, withAlignRight
     , renderElement
     , FilterModel, fromModel
     , singleTextFilter, multiTextFilter, singleDateFilter, rangeDateFilter, periodDateFilter, radioFilter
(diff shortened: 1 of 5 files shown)


src/UI/Filter.elm
```diff
@@ -232,6 +233,13 @@ withWidth newWidth (Filter filter) =
     Filter { filter | width = newWidth }
 
 
+{-| Aligns the filter to the right.
```
kress95, line 236: TODO: improve

PedroHLC (approved):
Looks good to me, including that brief description in the docs.

niqt (approved):

PR #285, opened 2021-08-22
Update README.md

VladimirLogachev:
#### :thinking: What?

Describe limitations for potential users in the README file.

#### :man_shrugging: Why?

To make potential users understand the limitations clearly before they even try to use the library so that they could save time.

#### :pushpin: Jira Issue

Not tracked.

#### :no_good: Blocked by

Nothing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,10 @@ The purpose of using Elm UI is mostly to keep components accessible.
 
 The usage of this library is mostly internal, but we are open to suggestions and feedback.
 
+## Limitations
+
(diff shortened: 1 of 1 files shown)


renanpvaz:
Hey @VladimirLogachev, I appreciate your intent with the PR but this package's purpose is to build internal UIs at Paack using it's design system, that's why you can't change the palette. If you think that it is unclear we can update the readme to emphasize that. Still if you want to, feel free use it as is or fork it and change the palette, we don't mind!

PR #292, opened 2021-09-02
S4S-110 | Add error text below TextField if it applies

segmentationfaulter:
#### :thinking: What?

Show error text below a text field if it applies

input-error

#### :man_shrugging: Why?

Since it was missing

#### :pushpin: Jira Issue

[S4S-110](https://paacklogistics.atlassian.net/browse/S4S-110)

### DIFF
--- a/src/UI/TextField.elm
+++ b/src/UI/TextField.elm
@@ -467,26 +467,21 @@ renderElement cfg (TextField prop opt) =
                 ContentSinglelineText ->
                     inputAnyOptions cfg msg prop opt
                         |> Input.text elAttrs
-                        |> textFieldError cfg opt.errorCaption
 
                 ContentMultilineText ->
                     inputMultilineOptions cfg msg prop opt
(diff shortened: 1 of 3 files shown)


miguelcoba (approved):

niqt (approved):

kress95 (approved):

src/UI/TextField.elm
```diff
@@ -492,6 +495,7 @@ renderElement cfg (TextField prop opt) =
                 ContentSpellChecked ->
                     inputAnyOptions cfg msg prop opt
                         |> Input.spellChecked elAttrs
+                        |> textFieldError cfg opt.errorCaption
```
renanpvaz, line 498: since you're adding it to every case doesn't it make sense to add it before the pattern match or where `nonStatic` is called?

renanpvaz (approved):

PedroHLC (approved):
Don't forget to bump the version before merging...

PR #297, opened 2021-09-10
Adds blur handler in focus config

segmentationfaulter:
#### :thinking: What?

Adds blur handler in focus config

#### :man_shrugging: Why?

Need to show [some suggestion](https://www.figma.com/file/GUqevGHHTKTnZT4VgaXfns/Communications?node-id=900%3A6294) under the `TextField` depending upon the focus status.

#### :pushpin: Jira Issue

Not tracked.

PR #298, opened 2021-09-10
Paack-UI V7

PedroHLC:
#### :thinking: What?

1. Kill ListView, SummaryItem, and SplitSelectable;
2. Move V2 components as the only ones;
3. Remove "Success Button";
4. Add generic colors, specified in Figma;
5. Rename "Effect" to "Effects";
6. Opaque "Paginator" and "Utils.Focus";
7. Renames to "NavigationContainer", plus, more opaque types;
8. Verbose names in Radio's sizes;
9. `Element.Decoration` instead of `Element.Attributes` in Palette.

#### :man_shrugging: Why?

- For (1): These components are too different across applications, it's too complex to have a generic component covering all the possibilities;
- For (2): Palette wasn't backwards-compatible; but we don't use V1 dialog anymore. Besides, were these "components versions" or "entire paack-ui versions";
- For (3): We never used it, and UX team decided to drop it;
- For (4): UX team had it, we use it in backgrounds;
- For (5): To match all of our other projects;
- For (9): This allows us to use it in "mouse over" transitions.

#### :pushpin: Jira Issue

Not tracked.

#### :no_good: Blocked by

#297 

#### :clipboard: Pending

Finish it.


PedroHLC:
Besides all this, I'm thinking about what we can do with the texts (for composition and width-shrink), and if there is some better way to fix the weird radio-button buggy focus behavior...

segmentationfaulter (approved):

src/UI/Page.elm
```diff
@@ -1,19 +1,18 @@
-module UI.NavigationContainer exposing
+module UI.Page exposing
```
renanpvaz, line 1: 🤗

elm.json
```diff
@@ -39,8 +35,7 @@
         "UI.Utils.Element",
         "UI.Utils.Focus",
         "UI.Utils.TypeNumbers",
-        "UI.V2.Dialog",
-        "UI.Layout.SplitSelectable"
+        "UI.V2.Dialog"
```
renanpvaz, line 38: If you're killing v2 components shouldn't this be just `UI.Dialog`?

renanpvaz (approved):

kress95 (approved):

miguelcoba (approved):

PR #301, opened 2021-09-17
S4S-133 | If an error caption is empty, it shouldn't make TextField borders red

segmentationfaulter:
#### :thinking: What?

Stops making borders red if error caption is an empty string

#### :man_shrugging: Why?

I had an `error` filed in my model of type `Maybe String`, for `Nothing` values of `error`, I was seeing red borders. Here is a code sample:

```elm
TextField.singlelineText Msg "Label" "value"
    |> TextField.withError (error |> Maybe.map identity |> Maybe.withDefault ""
```

#### :pushpin: Jira Issue

https://paacklogistics.atlassian.net/browse/S4S-133

### DIFF
--- a/src/UI/TextField.elm
+++ b/src/UI/TextField.elm
@@ -649,7 +649,9 @@ attrs : RenderConfig -> Properties msg -> Options msg -> List (Attribute msg)
 attrs cfg prop opt =
     let
         hasError =
-            opt.errorCaption /= Nothing
+            opt.errorCaption
+                |> Maybe.map showError
+                |> Maybe.withDefault False
 
         isPlaceholder =
             prop.currentValue == ""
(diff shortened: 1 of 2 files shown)


src/UI/TextField.elm
```diff
@@ -793,15 +795,26 @@ textFieldError : RenderConfig -> Maybe String -> Element msg -> Element msg
 textFieldError cfg errorCaption inputElement =
     case errorCaption of
         Just caption ->
-            Element.column
```
PedroHLC, line 796: What if, instead of adding the case here, in the `withError` you make it back to `Nothing` if the string is empty...

PedroHLC (approved):

miguelcoba (approved):

PedroHLC (approved):

kress95 (approved):

PR #314, opened 2021-10-04
Fix loader

sufiyanyusuf:
#### :thinking: What?

- Updated loader color to match updated palette
- Increased thickness for added prominence

#### :man_shrugging: Why?

- Let's put some ❤️in the loader

#### :pushpin: Jira Issue

Not tracked.

#### :no_good: Blocked by

Nothing.

#### :clipboard: Pending

- [ ] Add a determinate version ?

### :fire: Extra

### DIFF
--- a/src/UI/LoadingView.elm
+++ b/src/UI/LoadingView.elm
@@ -15,6 +15,7 @@ import Html exposing (Html)
 import Svg exposing (..)
 import Svg.Attributes exposing (..)
 import UI.Internal.Svg as Svg
+import UI.Palette as Palette
 
 
 {-| The version with a small (18x18px) spinner.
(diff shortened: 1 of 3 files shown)


src/UI/LoadingView.elm
```diff
@@ -58,17 +59,17 @@ spinnerSvg floatSize =
     svg
         [ width size
         , height size
-        , viewBox "0 0 38 38"
-        , stroke "#1747CE"
+        , viewBox "-1 -1 40 40"
```
PedroHLC, line 62: I see your padding hacks here :rofl:

PedroHLC (approved):

niqt (approved):

PR #316, opened 2021-10-08
Legacy sidebar

niqt:
#### :thinking: What?

Added legacy sidebar and legacy icons

#### :man_shrugging: Why?

Rems use a legacy sidebar

#### :pushpin: Jira Issue

 Not tracked.

#### :no_good: Blocked by

 Nothing.

### :fire: Extra

[REMS-CODE] (https://paacklogistics.atlassian.net/jira/software/projects/S3/boards/53?selectedIssue=S3-252)


src/UI/Document.elm
```diff
@@ -288,6 +291,14 @@ withSidebarStyle style (Document nav) =
     Document { nav | sidebarStyle = style }
 
 
+{-| `Document.withSidebarStyle` takes a `SidebarStyle` setting the
+appearance/behavior of the sidebar when it is enabled by the `hasMenu` flag.
+-}
+withLegacyTransitionMenuStyle : Document page msg -> Document page msg
```
kress95, line 297: I think changing `SidebarStyle` into 

```elm
type SidebarStyle
    = SidebarPersistent
    | SidebarNonPersistent
    | SidebarLegacyMigration
```

would be a better approach

kress95 (dismissed):

kress95:
and also preferer rebases to merges, we try to keep `paack-ui`'s history as linear as possible

src/UI/Document.elm
niqt: In this way i can't have Legacy and Persistent

src/UI/Document.elm
```diff
@@ -550,7 +561,14 @@ toBrowserDocument cfg pageItem (Document model) =
...
             if hasMenu then
-                if RenderConfig.isMobile cfg then
+                if RenderConfig.isMobile cfg && model.hasLegacySidebar then
+                    LegacySideBar.mobile cfg
+                        contentBody
+                        model.menu
+                        seenTitle
+                        maybeStack
+
+                else if RenderConfig.isMobile cfg && model.hasLegacySidebar == False then
```
PedroHLC, line 571: I'm not cool with checking the same condition twice. Do it as nested if...

src/UI/Document.elm
```diff
@@ -560,10 +578,18 @@ toBrowserDocument cfg pageItem (Document model) =
...
+                            if model.hasLegacySidebar then
+                                LegacySideBar.desktopPersistent cfg contentBody model.menu
+
+                            else
+                                SideBar.desktopPersistent cfg contentBody model.menu
 
                         SidebarNonPersistent ->
-                            SideBar.desktopNonPersistent cfg contentBody model.menu
+                            if model.hasLegacySidebar then
+                                Element.none
```
PedroHLC, line 589: This way we have an `Element.none`,  is this on purpose?

If not, then we should make legacy always be persistent...

src/UI/Document.elm
```diff
@@ -518,6 +528,7 @@ document applier (Model state) pagesPages =
         DocumentRecord pagesPages
             (menu applier state)
             SidebarPersistent
+            False
```
PedroHLC, line 531: This became confusing now, I recommend:

```suggestion
        { container = pagesPages
        , menu : = menu applier state
        , sidebarStyle = SidebarPersistent
        , hasLegacySidebar = False
        }
```

PR #319, opened 2021-10-12
Experiment with dropdown sizes

gregziegan:
#### :thinking: What?

I'd like to use multiple sizes for a dropdown. 

This is unfinished, I just wanted to see if others would like this and if i should continue.

#### :man_shrugging: Why?

I'm attempting to build a Split Button ([example](https://garden.zendesk.com/components/split-button)). I can use `Paack-Eng/elm-ui-dropdown` to accomplish this, but i thought this work could be useful within the design system.

#### :pushpin: Jira Issue

None

#### :no_good: Blocked by

None

### :fire: Extra


PedroHLC:
Closing this, it looks stale. I'll reopen+rebase if we'll ever come to need this

PR #331, opened 2021-11-18
Buttons update

sufiyanyusuf:
#### :thinking: What?

Switched on state for buttons

#### :man_shrugging: Why?

For use cases where we need to show something in a 'switched on ' state

#### :pushpin: Jira Issue

DES-127

#### :no_good: Blocked by

Nothing.

#### :clipboard: Pending

Nothing

### :fire: Extra


PedroHLC (approved):
This one already looks good to me!

kress95 (approved):

niqt (approved):

PedroHLC (changes requested):
You'll need to run `elm bump`, and add `switchedOn` to an `@docs` line.

segmentationfaulter (approved):

kress95 (commented):
Just fix what @PedroHLC asked, otherwise seems alright.

PedroHLC (approved):

PR #332, opened 2021-11-19
S4 center dialog

niqt:
#### :thinking: What?

Dialog in the center of the screen for application that doesn't use 100vh

#### :man_shrugging: Why?

Required from Squad 4

#### :pushpin: Jira Issue

 Not tracked.

#### :no_good: Blocked by

Nothing.

### DIFF
--- a/src/UI/Internal/Dialog.elm
+++ b/src/UI/Internal/Dialog.elm
@@ -128,8 +128,6 @@ mobileView : RenderConfig -> Dialog.Dialog msg -> Element msg
 mobileView cfg (Dialog.Dialog { title, icon } { body, buttons }) =
     Element.column
         [ mainBackground
-        , Element.centerX
-        , Element.centerY
         , Element.padding 32
         , Element.spacing 8
         , Border.rounded 6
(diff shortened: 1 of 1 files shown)


src/UI/Internal/Dialog.elm
```diff
@@ -68,6 +67,12 @@ desktopDialogView cfg (Dialog.Dialog { title, icon } { body, buttons }) =
         [ body
         , buttonsRow cfg buttons
         ]
+        |> Element.el
+            [ Element.centerX
+            , htmlAttribute (Html.Attributes.style "position" "fixed")
+            , htmlAttribute (Html.Attributes.style "top" "50%")
+            , htmlAttribute (Html.Attributes.style "bottom" "50%")
```
PedroHLC, line 74: Does it not work when put in the `Element.column` attributes?

niqt: No, has different behaviour

PedroHLC (approved):
This one looks good for me, easier to understand what's going on.

PedroHLC:
Like on #333, you'll need to bump the version...

PedroHLC:
Also, this broke the storybook...

PR #333, opened 2021-11-22
S4S-299 | Fix dialog vertical centering

segmentationfaulter:
#### :thinking: What?

Vertically centers dialog irrespective of the height of the page it is rendered in

#### :man_shrugging: Why?

Dialog was not centered if the contents of the page had a height more than `100vh`, Sergio reported it [here](https://paack.slack.com/archives/C020YT55GNB/p1637222453003000).

#### :pushpin: Jira Issue

[S4S-299](https://paacklogistics.atlassian.net/browse/S4S-299)

### :fire: Extra

@niqt approached this problem using css positions, he opened #332 to fix this problem. I tested this approach later and discussed it with @niqt, and we decided to open two different PR's, and let the team decide which way to go. Thanks

### DIFF
--- a/elm.json
+++ b/elm.json
@@ -3,7 +3,7 @@
     "name": "PaackEng/paack-ui",
     "summary": "Paack's Design System applied over Elm UI",
     "license": "BSD-3-Clause",
-    "version": "7.8.1",
+    "version": "7.8.2",
     "exposed-modules": [
         "UI.Alert",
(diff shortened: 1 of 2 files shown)


PedroHLC (commented):
This one hacks the body once a dialog is visible right? Does this mean the user's scrolling is lost?

PR #340, opened 2021-12-31
WS-544 | LMO-1407 | Add dropdown `Menu` component

kress95:
#### :thinking: What?

Ports `PopupMenu` from `wms-web` into a more stable `Menu` component.

#### :man_shrugging: Why?

Many apps are using context menu buttons, this component should fit all these uses.

#### :pushpin: Jira Issue

[WS-544](https://paacklogistics.atlassian.net/browse/WS-544)
[LMO-1407](https://paacklogistics.atlassian.net/browse/LMO-1407)

#### :no_good: Blocked by

Nothing.

#### :clipboard: Pending

- Nothing

### :fire: Extra

Screenshots

image

image

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-* @miguelcoba @PedroHLC @renanpvaz @segmentationfaulter @kress95 @niqt
+* @PedroHLC @segmentationfaulter @kress95 @niqt @sufiyanyusuf
--- a/src/UI/Menu.elm
+++ b/src/UI/Menu.elm
@@ -15,9 +15,8 @@ A menu can be created and rendered as in the following pipeline:
     Button.cmd ToggleMenu Button.primary
         |> Menu.menu ToggleMenu
             [ Menu.item Download
-                { icon = Icon.download
-                , title = "Download"
-                }
+                Icon.download
+                "Download"
             ]
         |> Menu.renderElement renderConfig
 
@@ -80,7 +79,7 @@ type MenuItem msg
 
 type alias InternalMenuItem msg =
     { color : Maybe Color
-    , icon : Maybe (String -> Icon)
+    , icon : String -> Icon
     , title : String
     , onClick : msg
     }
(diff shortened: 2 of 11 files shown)


PedroHLC (changes requested):

niqt (approved):

src/UI/Menu.elm
```diff
@@ -76,24 +93,32 @@ type alias InternalMenuItem msg =
...
-    msg
-    ->
-        { icon : Maybe (String -> Icon)
-        , title : String
-        , color : Maybe Color
-        }
-    -> MenuItem msg
-item onToggle { icon, title, color } =
-    MenuItem { onClick = onToggle, icon = icon, title = title, color = color }
+item : msg -> { icon : Maybe (String -> Icon), title : String } -> MenuItem msg
```
PedroHLC, line 101: 1 - I don't really like this record here, the types are very different for any confusion.
2 - The icon is a  maybe? it seems different from the documentation. If it's a maybe, then let's move it to with-pattern too...

PedroHLC (changes requested):

src/UI/Menu.elm
kress95: 1) I think it's better as a record, but it could be plain arguments as well

2) Never had to use it as a maybe, I was being overly careful with options here :/

PedroHLC (changes requested):
The API looks good, can you add a very dummy page to the storybook and deploy it?

PedroHLC:
Also, could you kindly update the CODEOWNERS?

antoniofents (approved):

sufiyanyusuf (changes requested):
Great work @kress95 , just a couple of points :

- Instead of having "item with color", we need a danger variant (lets be strict to avoid misuse). This with danger menu item should have its own Red/200 hover color
- Icon should be optional

.github/CODEOWNERS
```diff
@@ -1 +1 @@
-* @miguelcoba @PedroHLC @renanpvaz @segmentationfaulter @kress95 @niqt
+* @PedroHLC @segmentationfaulter @kress95 @niqt @Sufiyan
```
PedroHLC, line 1:
```suggestion
* @PedroHLC @segmentationfaulter @kress95 @niqt @sufiyanyusuf
```

PedroHLC (approved):
There is what sufiyan said. Everything else looks good to me!

sufiyanyusuf (approved):

PedroHLC (approved):

PR #341, opened 2022-01-05
[DELAYED] WS-544 | Improve menu API

chrisfls:
#### :thinking: What?

Remove icon from item function and add a with pattern function for it.

#### :man_shrugging: Why?

Just forgot that maybes should be put in with pattern 🙃 

#### :pushpin: Jira Issue

[WS-544](https://paacklogistics.atlassian.net/browse/WS-544)

#### :no_good: Blocked by

Nothing.

#### :clipboard: Pending

- Nothing

### :fire: Extra

Nothing

### DIFF
--- a/elm.json
+++ b/elm.json
@@ -3,7 +3,7 @@
     "name": "PaackEng/paack-ui",
     "summary": "Paack's Design System applied over Elm UI",
     "license": "BSD-3-Clause",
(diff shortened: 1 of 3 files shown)


niqt (approved):

PedroHLC:
No, let's save this change for a future major bump...

PR #347, opened 2022-01-26
DES-129 | LMO-1429 | Introduce the date-picker component

niqt:
#### :thinking: What?

Datepicker

#### :man_shrugging: Why?

Improve the user experience

#### :pushpin: Jira Issue

[DES-129](https://paacklogistics.atlassian.net/browse/DES-129) and [LMO-1429](https://paacklogistics.atlassian.net/browse/LMO-1429).

#### :no_good: Blocked by

Nothing.

#### :clipboard: Pending

Nothing.

PR #348, opened 2022-01-28
S4S-377 | Add withId for button component

segmentationfaulter:
#### :thinking: What?

Adds the ability to add an ID to button component

#### :man_shrugging: Why?

Our squad [need it](https://paacklogistics.atlassian.net/browse/S4S-377) to track clicks on buttons in Timeslot-change page through Google Tag Manager. Here is a small [slack conversation](https://paack.slack.com/archives/CUKQCEN30/p1643367475526139) to have more context.

#### :pushpin: Jira Issue

Not tracked.

### DIFF
--- a/src/UI/Button.elm
+++ b/src/UI/Button.elm
@@ -487,7 +487,6 @@ toggleView cfg hint toggleMsg current { size, id } =
         attrs =
             Primitives.roundedBorders size
                 :: (Element.onIndividualClick <| toggleMsg (not current))
-                :: (id |> Maybe.map ElementUtils.id |> Maybe.withDefault (Primitives.roundedBorders size))
                 :: (ARIA.toElementAttributes <| ARIA.roleToggleButton current)
                 ++ toggleTheme current
                 ++ iconLayout hint size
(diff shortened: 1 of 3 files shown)


src/UI/Button.elm
```diff
@@ -145,6 +152,7 @@ defaultOptions : Options
 defaultOptions =
     { width = WidthShrink
     , size = Size.default
+    , id = ""
```
PedroHLC, line 155: I think a Maybe would be better. Otherwise we'll have lots of empty strings in the model and I'm not sure how much overhead that will generate for the JS VM.

PedroHLC (changes requested):

PedroHLC:
Let's merge this one before #347

src/UI/Button.elm
kress95: the overhead is lesser for the empty string than it is for the maybes, but the id should be optional and the best way to represent that is using a maybe

src/UI/Button.elm
```diff
@@ -492,6 +515,7 @@ hyperlinkView cfg size width body action =
                 :: Font.regular
                 :: Font.underline
                 :: Element.pointer
+                :: (id |> Maybe.map ElementUtils.id |> Maybe.withDefault Element.pointer)
```
segmentationfaulter, line 518: @PedroHLC is there a better way of doing this?

PedroHLC: https://github.com/PaackEng/paack-ui/blob/15f172d45eec94c0eeff0568c4b183b293403859/src/UI/Internal/Basics.elm#L5

PedroHLC (approved):

kress95 (approved):

PR #357, opened 2022-02-16
WS-564 | Paginator scroll

kress95:
#### :thinking: What?

- Adds `withHeight` to the stateful table.
- Adds a a scrollbar to the stateful tables that have a paginator.

#### :man_shrugging: Why?

Fixes a design issue on wms.

It unblocks first pending point on https://github.com/PaackEng/wms-web/pull/630

I don't think solving the second point is worth for now.

#### :pushpin: Jira Issue

[WS-564](https://paacklogistics.atlassian.net/browse/WS-564)

#### :no_good: Blocked by

Nothing.

#### :clipboard: Pending

Nothing.

### :fire: Extra

Already up and running on https://paackeng.github.io/paack-ui/#Complex%20components/Tables/Desktop

### DIFF
--- a/src/UI/Tables/Stateful.elm
+++ b/src/UI/Tables/Stateful.elm
@@ -1387,21 +1387,15 @@ desktopView renderConfig prop opt =
         items =
             viewGetItems state opt
 
-        ( items_, footer ) =
+        items_ =
             case state.paginator of
-                Just ({ from, by } as paginator) ->
-                    ( items
+                Just { from, by } ->
+                    items
                         |> List.drop from
                         |> List.take by
-                    , paginator
-                        |> viewPaginator renderConfig (List.length items)
-                        |> Element.map prop.toExternalMsg
-                        |> Tuple.pair "paginator"
-                        |> List.singleton
-                    )
 
                 Nothing ->
-                    ( items, [] )
+                    items
 
         rows =
             List.map (rowWithSelection renderConfig prop.toExternalMsg state prop.toRow columns) items_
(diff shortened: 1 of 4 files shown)


PedroHLC (approved):
I'll never like a `withHeight`, but I don't know better solutions...

src/UI/Tables/Stateful.elm
```diff
@@ -1421,12 +1429,40 @@ desktopView renderConfig prop opt =
...
+                , Element.paddingEach padding
+                ]
+                [ Keyed.column
+                    [ Element.spacing 2
+                    , Element.height Element.fill
+                    , Element.width Element.fill
+                    , Element.alignTop
+                    , Element.scrollbars
+                    ]
+                    (headers :: rows)
```
PedroHLC, line 1446: should the headers scroll?

PedroHLC (approved):

kress95:
> I'll never like a `withHeight`, but I don't know better solutions...

it currently can work without it, but if I push the "headers scroll" feature it will be required

src/UI/Tables/Stateful.elm
kress95: as we chatted about in slack, fixing the header on top and making it scroll horizontally with the content has some usability issues if we do it the usual way (the vertical scrollbar ends on the right side of the content, forcing the user to scroll horizontally to reach it)

there are hacky ways to sync the header's position with the horizontal scrollbar, but doing it without js will cause performance issues, so it's better if I postpone it (and paack-ui has no js, so I'll be "breaking" a good package for that :/)

